Issue #1, opened 2020-05-19
Update the ZIO SQL project with the latest ZIO gitter8 template


jdegoes:
This project is using an older template and should be updated to work with the latest gitter8 template, available here:

<https://github.com/zio/zio-project-seed.g8>

jdegoes:
Closed by #24

PR #24, opened 2020-05-19
upgrade sbt build using latest g8 ZIO template

landlockedsurfer:
I've run g8 using the template given in the [link](https://github.com/zio/zio-project-seed.g8) given in issue #1 to generate a zio-sql project using the latest g8 template and copied over the generated files rather mechanically.
I'm not sure if we should keep the JS crossbuild?
Hope this pull request is a useful start to get this issue closed. :-)


jdegoes:
Yes, I think we should keep the JS crossbuild. Thank you for completing this!

PR #26, opened 2020-05-23
Simplify Type Signatures in FunctionDef

adamgfraser:
I think we can simplify the signature of `narrow` in `FunctionDef` from:

```scala
def narrow[A1 <: A, C](implicit ev: A1 =:= C): FunctionDef[C, B]
```

to:

```scala
def narrow[C](implicit ev: C <:< A): FunctionDef[C, B]
```

This also allows us to simplify the type signature of the `FunctionCallN` variants to eliminate the `A1` type parameter and only require implicit evidence that `(P1, P2) <:< A`, which should improve type inference.

Other changes just from running `fmt`.

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -106,11 +106,9 @@ object BuildHelper {
     }
 
   def platformSpecificSources(platform: String, conf: String, baseDirectory: File)(versions: String*) =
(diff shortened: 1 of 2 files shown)


jdegoes:
Love this change!

PR #27, opened 2020-05-23
Make Expr Covariant on Output Type

adamgfraser:
By moving `as` and `in` to implicit syntax we can recover covariance on output types of `Expr` and `FunctionDef`.

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -583,9 +583,9 @@ trait Sql {
     }
   }
 
-  sealed case class AggregationDef[-A, B](name: FunctionName) { self =>
+  sealed case class AggregationDef[-A, +B](name: FunctionName) { self =>
 
-    def apply[F, Source, A1 <: A](expr: Expr[F, Source, A1]): Expr[Features.Aggregated[F], Source, B] =
+    def apply[F, Source](expr: Expr[F, Source, A]): Expr[Features.Aggregated[F], Source, B] =
       Expr.AggregationCall(expr, self)
   }
 
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -65,7 +65,7 @@ trait Sql {
     type ColumnsRepr[T]
     type Append[That <: ColumnSet] <: ColumnSet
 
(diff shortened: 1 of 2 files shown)


adamgfraser:
Ah I see we want to make it invariant to prevent the type from being widened when we do set in an update operation. Could we create an Equal type class to prevent that? Since our type system is closed it seems like we could do it like we did with `IsNumeric`.

jdegoes:
Right, the widening lets us "set" absurd things. If we can fix that in a sound way (by leveraging our closed type system), then I'd certainly prefer covariance.

adamgfraser:
Let me work on that.

adamgfraser:
I think we can just require a `TypeTag` for `Value` in `set`.

adamgfraser:
This seems to be working now. If you try to get set a nonsensical value you get a compilation error that a type tag could not be found for any. If we wanted we could wrap this up in a trait to provide a more helpful error message.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -236,7 +236,7 @@ trait Sql {
   // WHERE baz > buzz
   sealed case class Update[A](table: Table.Aux[A], set: List[Set[_, A]], whereExpr: Expr[_, A, Boolean]) {
 
-    def set[F: Features.IsSource, Value](lhs: Expr[F, A, Value], rhs: Expr[_, A, Value]): Update[A] =
+    def set[F: Features.IsSource, Value: TypeTag](lhs: Expr[F, A, Value], rhs: Expr[_, A, Value]): Update[A] =
```
jdegoes, line 239: I'd also add a `def typeTag: TypeTag[Value]` member to `Update`, because we are requiring it here but not storing it inside the `Update` structure.

adamgfraser: Don't we need to capture it in `Set` since `Update` could contain zero or many sets?

jdegoes: Ah right, sorry. 👍

jdegoes: Ah right, sorry. 👍

jdegoes:
Thank you for this! Really like going back to covariant on the expression type.

Issue #6, opened 2020-05-19
Add AggregationDef for aggregations supported by ANSI-92


jdegoes:
We should support the full set of standard aggregations, such as `COUNT`, `SUM`, etc, which are defined by ANSI-92 and / or supported by Postgres, MySQL, SQL Server, and Oracle.

These would be added as `AggregationDef` values, in the companion object of `AggregationDef`.

If anything is unclear, please ask for more details!

jczuchnowski:
Looks like this is done. If there are more missing functions we can do it in another Issue/PR.

PR #29, opened 2020-05-25
Add missing aggregation functions

marekklis:
For #6 

@jdegoes is it enough?

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -595,6 +595,9 @@ trait Sql {
     val Count: AggregationDef[Any, Long] = AggregationDef(FunctionName("count"))
     val Sum                              = AggregationDef[Double, Double](FunctionName("sum"))
     val Arbitrary                        = AggregationDef[Any, Any](FunctionName("arbitrary"))
+    val Avg                              = AggregationDef[Double, Double](FunctionName("avg"))
+    val Min                              = AggregationDef[Any, Any](FunctionName("min"))
+    val Max                              = AggregationDef[Any, Any](FunctionName("max"))
   }
 
   sealed case class FunctionDef[-A, +B](name: FunctionName) { self =>


jdegoes:
@marekklis Fantastic start!

I'd look [here](https://www.oreilly.com/library/view/sql-in-a/9780596155322/ch04s02.html) and add any that are supported on all the latest versions of: MySQL, Postgres, SQL Server, and Oracle.

jczuchnowski:
I see later specs (2011) define also 
- STDDEV_POP
- STDDEV_SAMP
- VAR_SAMP
- VAR_POP
And both PostgreSQL and MS SQL support some form of it. I guess we could add these if MySQL and Oracle also support them.

marekklis:
I thought that this issue is about adding aggregate functions defined by ANSI-92 standard. Based on [this](https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) doc these functions are: `count`, `sum`, `avg`, `min`, `max`. 
@jdegoes @jczuchnowski I'll add more functions you suggested

jdegoes:
I think ANSI-92 should be the guide, but if some function is supported by all 4 databases, then it's safe for inclusion, even if it does _not_ appear in any ANSI standard. Meanwhile, if something appears even in ANSI-92, but it is not supported by one of the 4, it should not appear in the `Sql` module.

marekklis:
Based on the following docs:

- https://dev.mysql.com/doc/refman/8.0/en/group-by-functions.html
- https://www.postgresql.org/docs/current/functions-aggregate.html
- https://docs.microsoft.com/en-us/sql/t-sql/functions/aggregate-functions-transact-sql?view=sql-server-ver15
- https://docs.oracle.com/database/121/SQLRF/functions003.htm#SQLRF20035
aggregation functions supported by all 4 servers (mysql, postgresql, sql sever and oracle) are:

1. `AVG()`	     
2. `COUNT()`      
3. `MAX()`	     
4. `MIN()`	     
5. `SUM()`
6. `STDDEV()`	 (`STDEV` for SQL Server)
7. `STDDEV_POP()` (`STDEVP` for SQL Server)
8. `STDDEV_SAMP()` (`STDEV` for SQL Server)	     
9. `VAR_POP()`	 (`VARP` for SQL Server)
10. `VAR_SAMP()`   (`VAR` for SQL Server)
11. `VARIANCE()`	 (`VARP` for SQL Server)

marekklis:
@jdegoes Should I add `STDDEV_POP`, `STDDEV_SAMP`, `VAR_POP` and `VAR_SAMP` functions (even if the name of these functions are a bit different for SQL Server)?

jdegoes:
If the names are different, we should not include them here, but rather we can add them later in subtypes of `Sql`, e.g. `SqlServer extends Sql`, etc.

PR #30, opened 2020-05-25
Remove Extraneous Type Parameter from Expr.Source

adamgfraser:


adamgfraser:
That's a good point. Because `A` is invariant in `Source` right now if you don't specify a type parameter it will be inferred as `Nothing`, so maybe that is a good forcing function to require specifying a type parameter or constructing it in a way where the type can be inferred from the return type as in `mkColumns`.

jdegoes:
All right, will close this for now, we can re-open later if we want, after we prevent external construction of any of the terms.

Issue #5, opened 2020-05-19
Add FunctionDef for functions in ANSI-92


jdegoes:
We should support the full set of standard functions, which are defined by ANSI-92 and / or supported by Postgres, MySQL, SQL Server, and Oracle.

These would be added as `FunctionDef` values, in the companion object of `FunctionDef`.

If anything is unclear, please ask for more details!

jczuchnowski:
@marekklis is there anything more to do here?

marekklis:
I thought about conversion function but I checked that usage of that function is different for the different DB servers.
So, I think there is nothing to add.

PR #32, opened 2020-05-26
Add string functions

marekklis:
For #5 

Based on the following docs:
- https://dev.mysql.com/doc/refman/8.0/en/string-functions.html
- https://www.postgresql.org/docs/12/functions-string.html
- https://docs.microsoft.com/en-us/sql/t-sql/functions/string-functions-transact-sql?view=sql-server-ver15
- https://docs.oracle.com/database/121/SQLRF/functions002.htm#SQLRF51178

string functions supported by all 4 servers (mysql, postgresql, sql sever and oracle) are:
- `ASCII`	 
- `CONCAT`	 
- `LOWER`	 
- `LTRIM`	 
- `REPLACE`  
- `RTRIM`	 
- `TRIM`	 
- `UPPER`

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -651,11 +651,16 @@ trait Sql {
     val WidthBucket = FunctionDef[(Double, Double, Double, Int), Int](FunctionName("width bucket"))
 
     //string functions
+    val Ascii       = FunctionDef[String, Int](FunctionName("ascii"))
     val CharLength  = FunctionDef[String, Int](FunctionName("character length"))
+    val Concat      = FunctionDef[(String, String), String](FunctionName("concat"))
     val Lower       = FunctionDef[String, String](FunctionName("lower"))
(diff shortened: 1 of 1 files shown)


zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -651,11 +651,16 @@ trait Sql {
     val WidthBucket = FunctionDef[(Double, Double, Double, Int), Int](FunctionName("width bucket"))
 
     //string functions
+    val Ascii       = FunctionDef[String, Int](FunctionName("ascii"))
     val CharLength  = FunctionDef[String, Int](FunctionName("character length"))
+    val Concat      = FunctionDef[(String, String), String](FunctionName("concat"))
```
marekklis, line 656: I defined it as `FunctionDef[(String, String), String]` because Oracle version of `CONCAT` function is limited to only two input arguments

jdegoes:
Nice work! Thank you for the extensive research on this one!

PR #33, opened 2020-05-26
Add Transformation Functions for Expr

adamgfraser:
Initial work on moving to fixed point encoding for Expr and adding transformation functions.

@jdegoes Am I going in the right direction here?

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -465,7 +465,7 @@ trait Sql {
    * Models a function `A => B`.
    * SELECT product.price + 10
    */
-  sealed case class Expr[F, -A, +B](caseValue: Expr.ExprCase[F, A, B, Expr[_, _, _]]) { self =>
+  sealed case class Expr[F, -A, +B](caseValue: Expr.ExprCase[F, A, B, Expr]) { self =>
 
     def +[F2, A1 <: A, B1 >: B](that: Expr[F2, A1, B1])(implicit ev: IsNumeric[B1]): Expr[F :||: F2, A1, B1] =
       Expr.binary(self, that, BinaryOp.Add[B1]())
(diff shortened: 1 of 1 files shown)


zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -512,12 +512,36 @@ trait Sql {
...
+      caseValue match {
+        case c: Expr.AggregationCallCase[_, _, _, _, Expr[_, _, _]] =>
+          f(Expr.AggregationCallCase(c.param.fold(f), c.aggregation))
+        case c: Expr.BinaryCase[_, _, _, _, Expr[_, _, _]] =>
+          f(Expr.BinaryCase(c.left.fold(f), c.right.fold(f), c.op))
+        case c: Expr.FunctionCall1Case[_, _, _, _, Expr[_, _, _]] =>
+          f(Expr.FunctionCall1Case(c.param.fold(f), c.function))
+        case c: Expr.FunctionCall2Case[_, _, _, _, _, _, Expr[_, _, _]] =>
+          f(Expr.FunctionCall2Case(c.param1.fold(f), c.param2.fold(f), c.function))
+        case c: Expr.FunctionCall3Case[_, _, _, _, _, _, _, _, Expr[_, _, _] @unchecked] =>
```
adamgfraser, line 532: I get a warning here that the type is eliminated by erasure here but don't get any warnings with the variants with fewer type parameters so seemed like it is a compiler issue.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -545,50 +569,103 @@ trait Sql {
...
+
+    private[sql] def source[A, B](tableName: TableName, column: Column[B]): Expr[Features.Source, A, B] =
+      Expr(Expr.SourceCase(tableName, column))
+
+    trait ExprCase[F, -A, +B, +T]
+
+    sealed case class AggregationCallCase[F, A, B, Z, T](param: T, aggregation: AggregationDef[B, Z])
+        extends ExprCase[Features.Aggregated[F], A, Z, T]
+
+    sealed case class BinaryCase[F1, F2, A, B, T](left: T, right: T, op: BinaryOp[B])
```
jdegoes, line 630: I'm concerned that because of the need to use existentials in the recursive type (`Expr`), we have lost some type safety: we have no idea there is any connection between `T` and `B`.

(Similarly elsewhere, but this is a nice representative example.)

You can also see this happening not just with `B`, the output of the expression, but with features `F` too. 

I think to "solve" this problem we would have to make `T : [*, *] => *`, e.g. `T[F, B]` (assuming the source won't change in any recursive position — which may not be valid, need to think about that). This way here you could plug in `T[F1, B]` for left and `T[F2, B]` for right.

Then to regain fold, we could use `type Fold[F, B] = [Z] =>> Z`. 

Unfortunately the aesthetics / usability of this is rather terrible. Maybe this is a reason to not go down the recursion schemes path?

adamgfraser: What do you think of this? The internal implementation is fairly ugly but it seems like it creates a relatively nice API, at least given the complexity of the underlying data.

jdegoes:
My feeling looking at this:

> def fold[Z](f: Expr.ExprCase[_, Nothing, Any, ({ type Fold[F, -A, +B] = Z })#Fold] => Z): Z = {

Is that we've gone too far, and while experts will have no trouble understanding or using this, we have types like this inside the code base, it will discourage use and contributors. Although on that latter point, we are already discouraging contributors by using path-dependent types and type refinements.

Thoughts?

adamgfraser:
@jdegoes I think that is fair. There is a question in my mind of how much users will be working with this method directly at all. It seems like whether it is this or implementing a visitor to deal with a GADT with ten different cases, if users are having to do it then it is kind of a disaster.

So I think it comes down more to contributor accessibility, where I agree the visitor pattern would definitely be more accessible, though probably actually slightly less ergonomic if you know what you are doing. And then the consideration is just that there is already a lot of other complicated stuff in this library so I don't know if it is really moving the needle on contributor complexity.

Happy to change.

robmwalsh:
@adamgfraser closing because this is a little bit stale. Feel free to reopen if needed.

PR #34, opened 2020-05-29
Support selecting columns using existing column names

jgoday:
Implicit conversion from Expr[F, A, B] to Selection[F, A, SelectionSet.Cons[A, B, SelectionSet.Empty]]
to support select with default column names:

```scala
val userId :*: dob :*: fName :*: lName :*: _ = userTable.columns

val basicSelect = select { fName ++ lName } from userTable
```

Would it be possible to modify exprName to allow select arbitrary expressions with an auto generated name ?
```scala
  def exprName[F, A, B](expr: Expr[F, A, B]): Option[String] =
    expr match {
      case Expr.Source(_, c) => Some(c.name)
      case _ => None
    }
```

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -461,15 +461,6 @@ trait Sql {
 
   type :||:[A, B] = Features.Union[A, B]
 
-  def exprName[F, A, B](expr: Expr[F, A, B]): Option[String] =
-    expr match {
-      case Expr.Source(_, c) => Some(c.name)
-      case _ => None
-    }
-
-  implicit def expToSelection[F, A, B](expr: Expr[F, A, B]): Selection[F, A, SelectionSet.Cons[A, B, SelectionSet.Empty]] =
-    Selection.computedOption(expr, exprName(expr))
-
   /**
    * Models a function `A => B`.
    * SELECT product.price + 10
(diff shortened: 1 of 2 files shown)


jczuchnowski:
I just stumbled upon this problem. We need this.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -461,6 +461,15 @@ trait Sql {
 
   type :||:[A, B] = Features.Union[A, B]
 
+  def exprName[F, A, B](expr: Expr[F, A, B]): Option[String] =
+    expr match {
+      case Expr.Source(_, c) => Some(c.name)
+      case _ => None
+    }
+
+  implicit def expToSelection[F, A, B](expr: Expr[F, A, B]): Selection[F, A, SelectionSet.Cons[A, B, SelectionSet.Empty]] =
```
jdegoes, line 470: Can you place this into the companion object of `Selection`? It should be found there anyway, so you shouldn't need to import it, but it will keep it tidier than placing at the top level.

jdegoes:
@jgoday Excellent! Just a small hint for improving organization, then good to merge!

zio-sql/jvm/src/main/scala/Sql.scala
jgoday: Ok, no problem, but then it will be necessary to import Selection._ (see TestBasicSelect)
```scala
val userSql = new Sql { self =>
    import self.ColumnSet._
    import self.Selection._
...
```

jdegoes:
@jgoday You have a chance to revisit this? Would love to get this merged in!

jgoday:
> @jgoday You have a chance to revisit this? Would love to get this merged in!

Sorry @jdegoes, I have asked this last week ->

> Ok, no problem, but then it will be necessary to import Selection._ (see TestBasicSelect)
> 
> ```scala
> val userSql = new Sql { self =>
>     import self.ColumnSet._
>     import self.Selection._
> ...
> ```
> 
> @jdegoes Is this ok ?

jdegoes:
@jgoday I think we should try for zero imports required. Let me take another look at the right place to put this...

jdegoes:
@jgoday Can you try putting into the companion object of `Expr`? No imports should be required then.

jgoday:
> @jgoday Can you try putting into the companion object of `Expr`? No imports should be required then.

@jdegoes TestBasicSelect is working now without extra import clause

PR #35, opened 2020-05-29
Add current_timestamp function

marekklis:
There are many date&time functions but only `CURRENT_TIMESTAMP()` is supported by Postgres, MySQL, SQL Server and Oracle.

- https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html
- https://www.postgresql.org/docs/12/functions-datetime.html
- https://docs.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql?view=sql-server-ver15
- https://docs.oracle.com/database/121/SQLRF/functions002.htm#SQLRF51180

For #5


jdegoes:
Again, love the meticulous research! Thank you!

Issue #15, opened 2020-05-19
Create DELETE query examples


jdegoes:
We need a few examples of DELETE queries under different schemas, both to validate the design and flesh out edge cases, as well as to show people how to use the data type to model DELETE queries.

unclebob418:
I'll take this one and  #12 #13 #14 #16 - I'll need these for #8 and will probably come up with more edge cases from doing that. From watching the video I'm 90% sure I can write at least queries that will compile but won't be valid sql and vice versa. I also have some suspicions on group by.
`delete from a left join b on a.id = b.id where b.c = d` this isn't valid but I think will probably compile
`delete a from a left join b on a.id = b.id where b.c = d` this is valid but I don't think we can currently form this statement
guess I'll find out soon enough!

jdegoes:
Awesome, appreciate the identification of edge cases too!

jczuchnowski:
@robmwalsh can this issue be closed (and related mentioned by you)?

robmwalsh:
The reason it was originally left open was because the issue requested "under multiple schemas", but I'm not sure we need additional schemas - I think there's enough there to do most queries you're going to see in the wild.

jczuchnowski:
Ok, so I think I'll close these issues, as it's getting harder and harder to keep up with what is actually to do.
And yeah, I think the current examples are sufficient for now and we can expand them as needed.

Issue #14, opened 2020-05-19
Create SELECT join query examples


jdegoes:
We need a few examples of SELECT join queries under different schemas, both to validate the design and flesh out edge cases, as well as to show people how to use the data type to model SELECT join queries.

Issue #13, opened 2020-05-19
Create aggregation SELECT queries


jdegoes:
We need a few examples of SELECT aggregation queries under different schemas, both to validate the design and flesh out edge cases, as well as to show people how to use the data type to model SELECT aggregation queries.

Issue #12, opened 2020-05-19
Create simple SELECT query examples


jdegoes:
We need a few examples of SELECT query under different schemas, both to validate the design and flesh out edge cases, as well as to show people how to use the data type to model SELECT queries.

PR #40, opened 2020-05-31
WIP Examples

unclebob418:
I've created a schema and some data and started creating some queries. Not sure if I've put the .sql file in the right location but happy to move it. This is only for sql server at the moment.

I've been raising issues as I go and fixed a couple of minor things (like missing uuid type and bigdecimal not being "numeric")

From a usability point of view, so far it's been ok, but not spectacular. Error messages are helpful but frequently very very long due to the complexity of the types, so take a fair while to try to figure out the problem. Having to wrap a select in brackets 

One thing that's been a bit annoying is namespace issues, where different tables contain foreign keys for the same table. I've worked around this by stuffing things in objects, but I think it would work a lot better if the tables contained their column vals so you could write e.g. `users.userId` instead of awkward imports, but I'm not really sure how we could achieve this.


unclebob418:
relates to #12 #13 #14 #15, need to add one for 16

unclebob418:
@jdegoes or @adamgfraser can you please review?

zio-sql/sql/shop.sql
```diff
@@ -0,0 +1,199 @@
...
+    product_id uniqueidentifier not null,
+    quantity integer not null,
+    unit_price money not null
+)
+
+
+insert users
+    (usr_id, first_name, last_name, dob)
+values
+    ('60b01fc9-c902-4468-8d49-3c0f989def37', 'Ronald', 'Russell', '1983-01-05'),
```
jdegoes, line 51: Just checking this is all fake data and not previously owned test data IP, right? 😉

unclebob418: Wrote it all myself or randomly generated e.g. order details started life like this before I grouped them

```sql
insert into order_details
select top 5
    order_id,
    product_id,
    1 as quantity,
    price
from orders o cross apply(
select top 1 *
    from product_prices pp
    where pp.effective <= o.order_date
    order by effective desc, newid() ) a
order by newid()
```

jdegoes:
Awesome work and identified some issues we have to work on! Will comment on the issues separately (one of them, I don't think we need to 'fix', at least at the type level).

Thank you!

Issue #3, opened 2020-05-19
Add all binary operators supported by ANSI-92


jdegoes:
We should support the full set of binary operators, like `*`, `+`, etc., which are defined by ANSI-92 and / or supported by Postgres, MySQL, SQL Server, and Oracle.

These would be added as `BinaryOp` subtypes, with new methods defined in `Expr` like the methods for `*`, and `&&`.

If anything is unclear, please ask for more details!

jczuchnowski:
I'll work on this.

jdegoes:
@jczuchnowski Awesome, thank you!

jczuchnowski:
As I move forward I will be putting here information about decisions I make and findings realated to differences between the databases.
For starters:
- `%` as a modulo operator is not supported by Oracle
- `^` as an exponentiation operator is not supported by MySQL

PR #41, opened 2020-06-04
add binary operators

jczuchnowski:
Fix https://github.com/zio/zio-sql/issues/3

There's a lot more but every database does something differently.


jdegoes:
@jczuchnowski Awesome, thank you!

Issue #8, opened 2020-05-19
Add rendering to SQL for core data types


jdegoes:
We should add rendering to SQL for core data types, with several options:

```scala
sealed trait RenderMode
object RenderMode {
  case object Compact
  final case class Pretty(indent: Int)
}
...

// On all SQL types
def render(mode: RenderMode): String = /* calls render builder */

private[zio] def renderBuilder(builder: StringBuilder: mode: RenderMode): Unit
```

The use of `StringBuilder` is for performance reasons and can be hidden as a private implementation detail.

Note that extreme care should be taken when rendering strings into SQL, as they must be properly escaped, to avoid SQL injection attacks.

unclebob418:
I'll give this one a go

jdegoes:
Wow, that's awesome. Good luck and let me know if you need any help!

unclebob418:
I'll finish watching the video before I do too much on this. is the intention to return `PreparedStatement`s at the top level? I'd think ideally we don't execute queries that aren't parameterised.

jdegoes:
I would aim for a simple string rendering. Currently we don't have parameterized queries, which would be necessary to generate `PreparedStatement`. Also I think the core of the library should be usable without JDBC or even on Javascript.

We could consider this string rendering a step toward JDBC and also extremely useful for diagnostics / debugging / development.

jczuchnowski:
Closing this. @robmwalsh did a lot of work here. The rendering is now database specific.

PR #42, opened 2020-06-11
WIP Add Rendering

robmwalsh:
very naive attempt at #8, @jdegoes can you please review?
- no escaping for sql injection
- values not rendered correctly
- no group by
- no order by
- selects in joins aren't qualified by the table they're from (which is a problem if the same column name exists in both tables being joined)

issues discovered:
- aggregations go boom at runtime (unimplemented implicit defs)
- offset doesn't seem to be supported by sql server or oracle unless you use `fetch first/next`. In sql server this is considered part of `order by` and mysql doesn't support `fetch first`. I think we need a dialect as part of the signatures for `render` and `renderBuilder` to deal with dialect specific issues like top vs. limit vs. rownum (line 353)
- in an update statement, `set` isn't required to be non-empty. I've partially fixed this with `UpdateBuilder` but it's still possible to create an invalid `Update`

got the following output for the examples
```sql
select first_name as first, last_name as last from users
select first_name as first_name, last_name as last_name from users limit 3, 2
update users set first_name = Fred where true and first_name = Terrence
delete from users where first_name = Terrence
select first_name as first_name, last_name as last_name, order_date as order_date from users left join orders on usr_id = usr_id
```
need implementations for Aggregation implicit defs
```
Exception in thread "main" scala.NotImplementedError: an implementation is missing
	at scala.Predef$.$qmark$qmark$qmark(Predef.scala:347)
	at zio.sql.Sql$Features$IsAggregated$.AggregatedIsAggregated(Sql.scala:711)
```

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -1,7 +1,6 @@
 package zio.sql
 
 import scala.language.implicitConversions
-
 import java.time._
 import java.util.UUID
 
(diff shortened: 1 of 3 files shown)


jdegoes:
@unclebob418 Looks like the right track! Did you want to keep working on this, and keep it open, or merge now and fixup later?

The one thought I had: it would be quite useful to have a `SqlBuilder` data type that can be used to accumulate SQL strings, instead of a `StringBuilder`. This would allow, at least, a notion of indentation, which would be useful in pretty printing SQL strings.

robmwalsh:
@jdegoes RenderMode.Pretty has the concept of indentation, I'm just ignoring it at the moment. I think eventually an option for syntax highlighting may be nice but probably the most important thing would be getting something that can run against a db. I've added implementations for the aggregation defs but I think the rest of the issues I've noted above are still valid.

I'm not convinced having everything extend renderable is the correct approach. We don't yet have the concept of dialects but once we do I think this may be better represented in the individual dialect modules. Can we merge this for now so people have something to work with? cc @lemastero

Issue #43, opened 2020-06-13
need to be able to delete from joined tables


unclebob418:
split from #37 

All major dialects provide some way to delete from a joined table but the table to be deleted from needs to be explicitly provided. We could specify a "zio-sql" way of doing this (with dialect specific implementations) e.g. I'd suggest
```scala
delete(orders).from((orders join users).on(fkUserId === userId)).where(fName === "Fred")
```
or we could just disallow this and let the dialect specific modules introduce the concept of deleting from a join in a manner similar to the dialect of choice (e.g. using for postgres)

jdegoes:
I like the proposed syntax. We should support this and translate to dialect-specific variants in the rendering.

PR #44, opened 2020-06-13
Force delete from source

### LINKED ISSUE #37: force `deleteFrom` argument to be a source table
users may only delete from a source table, not a joined table.
```scala
val basicDelete = deleteFrom(users).where(fName === "Fred") //valid
val joinDelete = deleteFrom((orders join users).on(fkUserId === userId)).where(fName === "Fred") //not valid but compiles
```

All major dialects provide some way to delete from a joined table but the table to be deleted from needs to be explicitly provided. We could specify a "zio-sql" way of doing this (with dialect specific implementations) e.g. I'd suggest
```scala
delete(orders).from((orders join users).on(fkUserId === userId)).where(fName === "Fred")
```
or we could just disallow this and let the dialect specific modules introduce the concept of deleting from a join in a manner similar to the dialect of choice (e.g. `using` for postgres)

@jdegoes do you have any preference here?

unclebob418:
partially fixes #37 - requires `deleteFrom` to be a source table, doesn't address how we support deleting from a join; raised #43 for this

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -55,11 +55,11 @@ trait Sql {
     abstract class AbstractIsNumeric[A: TypeTag] extends IsNumeric[A] {
       def typeTag = implicitly[TypeTag[A]]
     }
-    implicit case object TShortIsNumeric  extends AbstractIsNumeric[Short]
-    implicit case object TIntIsNumeric    extends AbstractIsNumeric[Int]
-    implicit case object TLongIsNumeric   extends AbstractIsNumeric[Long]
-    implicit case object TFloatIsNumeric  extends AbstractIsNumeric[Float]
-    implicit case object TDoubleIsNumeric extends AbstractIsNumeric[Double]
+    implicit case object TShortIsNumeric      extends AbstractIsNumeric[Short]
+    implicit case object TIntIsNumeric        extends AbstractIsNumeric[Int]
+    implicit case object TLongIsNumeric       extends AbstractIsNumeric[Long]
+    implicit case object TFloatIsNumeric      extends AbstractIsNumeric[Float]
+    implicit case object TDoubleIsNumeric     extends AbstractIsNumeric[Double]
     implicit case object TBigDecimalIsNumeric extends AbstractIsNumeric[BigDecimal]
   }
 
@@ -191,6 +191,9 @@ trait Sql {
       val columns: F[TableType]
       val columnsUntyped: List[Column.Untyped]
     }
+    object Source {
+      type Aux[F[_], A, B] = Table.Source[F, A] { type TableType = B }
+    }
 
     sealed case class Joined[F, A, B](
       joinType: JoinType,
@@ -217,7 +220,7 @@ trait Sql {
   def select[F, A, B <: SelectionSet[A]](selection: Selection[F, A, B]): SelectBuilder[F, A, B] =
     SelectBuilder(selection)
 
-  def deleteFrom[A](table: Table.Aux[A]): DeleteBuilder[A] = DeleteBuilder(table)
+  def deleteFrom[F[_], A, B](table: Table.Source.Aux[F, A, B]): DeleteBuilder[F, A, B] = DeleteBuilder(table)
 
   def update[A](table: Table.Aux[A]): Update[A] = Update(table, Nil, true)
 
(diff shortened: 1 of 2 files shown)


zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -217,7 +220,7 @@ trait Sql {
   def select[F, A, B <: SelectionSet[A]](selection: Selection[F, A, B]): SelectBuilder[F, A, B] =
     SelectBuilder(selection)
 
-  def deleteFrom[A](table: Table.Aux[A]): DeleteBuilder[A] = DeleteBuilder(table)
+  def deleteFrom[F[_], A, B](table: Table.Source.Aux[F, A, B]): DeleteBuilder[F, A, B] = DeleteBuilder(table)
```
jdegoes, line 223: Could you try `table: Table.Source.Aux[_, A, _]` so we can avoid introducing `F` and `B`?

unclebob418: The first parameter of Table.Source.Aux is higher kinded, so can't just use _. @adamgfraser suggested using a type lambda which I think could work but figured we don't really care about it being a Table.Source.Aux past a certain point, so could just widen it to just be a Table.Aux[A] instead. Still have a couple of extra parameters on `DeleteBuilder` but it's an intermediate type so don't think this should be an issue - the `Delete` type still only has one parameter which I think was the goal.  What do you think of this approach?

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -185,11 +190,23 @@ trait Sql {
...
+      type Cols
       val name: TableName
-      val columnSchema: ColumnSchema[A]
-      val columns: F[TableType]
-      val columnsUntyped: List[Column.Untyped]
+      val columnSchema: ColumnSchema[Cols]
+      val columns: Repr[TableType]
+    }
+    object Source {
+      type Aux_[F[_], B] = Table.Source {
```
unclebob418, line 201: This is a bit ugly but couldn't think of another way to have an explicit return type for `Cons.table` (line 104)

unclebob418:
fixed source and tidied up the unused columnsUntyped issue (it wasn't defined for the `Table` trait). @jdegoes can you please review?

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -95,13 +95,16 @@ trait Sql {
     sealed case class Cons[A, B <: ColumnSet](head: Column[A], tail: B) extends ColumnSet { self =>
       type ColumnsRepr[T]            = (Expr[Features.Source, T, A], tail.ColumnsRepr[T])
       type Append[That <: ColumnSet] = Cons[A, tail.Append[That]]
+      type TableType
```
jdegoes, line 98: This can be deleted!

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -185,11 +190,23 @@ trait Sql {
...
-      val columnsUntyped: List[Column.Untyped]
+      val columnSchema: ColumnSchema[Cols]
+      val columns: Repr[TableType]
+    }
+    object Source {
+      type Aux_[F[_], B] = Table.Source {
+        type Repr[X] = F[X]
+        type Cols    = B
+      }
+      type Aux[F[_], A, B] = Table.Source {
```
jdegoes, line 205: This one is not used, so we can delete it, and rename `Aux_` => `Aux`.

unclebob418: @jdegoes `deleteFrom` (line 238) uses `Aux[F[_], A, B]` because it needs to know what `TableType` (aka `A`) is to pass through DeleteBuilder to ensure the `where` expression is valid.
`table` (line 104) uses `Aux_[F[_], B]` because it can't know what `TableType` (aka `A`) is for the return type.
Both are only used once, but I can't see a way to use the same `Aux` pattern for both.

zio-sql/jvm/src/main/scala/Sql.scala
jdegoes: @unclebob418 You don't need this `type TableType` anymore, right? So we can delete and then merge

unclebob418: fixed & merged changes from master - should be good to go now!

jdegoes:
Great work on this!

Issue #36, opened 2020-05-30
`in` should support sub queries that return a single column


unclebob418:
`select * from a where a.id in (select id from b)` is a fairly common pattern in sql which zio-sql doesn't currently support.

```scala
    val deleteFromWithSubquery = deleteFrom(orders).where(fkUserId in {
      select(userId as "id") from users where (fName === "Fred") //this should work
    })
```

```
type mismatch;
 found   : this.Read.Select[this.Features.Source,this.users.TableType,this.SelectionSet.Cons[this.users.TableType,java.util.UUID,this.SelectionSet.Empty]]
    (which expands to)  this.Read.Select[this.Features.Source,this.users.TableType,this.SelectionSet.Cons[this.users.TableType,java.util.UUID,this.SelectionSet.Empty.type]]
 required: this.Read[java.util.UUID] 
```

maciejbak85:
I can take it

PR #46, opened 2020-07-16
Allow `in` to accept single column subqueries

kitlangton:
Had a lovely time pairing with @unclebob418 today in an attempt to solve #36. I'm going to give a brief recap of the issue and why we ended up where we did.

Given the (previously) compile-error-inducing example `fkUserId.in(select(userId) from users where ...))`

There's the following problem: An `Expr[F,A,B]`'s `in` method expects a `Read[B]` parameterized by its own `B`, which would be an `Int` in this case.
```scala
def in[B1 >: B](set: Read[B1]): Expr[F, A, Boolean] = Expr.In(self, set)
```
While `Selection`'s B is actually going to be its entire `SelectionSet`:
```scala
sealed case class Select[F, A, B <: SelectionSet[A]](...) extends Read[B] { self =>
                               👆                  😭                 👆
```
Even if the selection set contains a single `Int`, `Select`'s `B` is going to be a `Cons[Int, Empty]`, which is certainly not an Int, and makes the compiler very sad.

The initial thought was to create a `Read.SingleColumnSelect` case class which would simply serve to wrap a Select of a single column, extending `Read[B]` where `B` matches its `Cons[B,Empty]` rather than the SelectionSet itself. If we did that, we could write an implicit conversion to wrap any such singleton `Select`s in this new type-transmogrifying wrapper and voila, the types would align!

Alas, it turns out that such an implicit conversion would only fire when we _explicitly_ typed the `Select` as `Read[Int]`, and would continue to fail otherwise—remaining the _wrong_ `Read[SelectionSet.Cons[...]`. We think this is because the `Select` is, after all, a valid `Read`, it merely has the wrong type parameter. Our thinking is that implicit conversions probably don't work this way, converting `Read[A]`s to `Read[B]`s without being explicitly asked to do so (in fact, things might become very unpredictable if they did).

So, even though this would've been a more straightforward solution, we responded to this obstacle by creating a _new_ type, our glorious friend `Inable` (also, definitely renamable). Basically, the same technique described above is used, only with this new type, such that the implicit conversions will trigger. 

We think it's worth reifying the constraint that "Only one expression can be specified in the select list when the subquery is not introduced with EXISTS", as SQL will tell you if you attempt to use `IN` with a multi-column subquery. But perhaps there's a more elegant approach. 🥰 Looking forward to working more on this after hearing your thoughts.

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -284,7 +284,6 @@ trait Sql {
   }
 
   object Read {
-
     sealed case class Select[F, A, B <: SelectionSet[A]](
       selection: Selection[F, A, B],
       table: Table.Aux[A],
@@ -317,7 +316,7 @@ trait Sql {
 
     sealed case class Literal[B: TypeTag](values: Iterable[B]) extends Read[B]
 
-    def lit[B: TypeTag](values: B*): Read[B] = Literal(values.toSeq)
+    def lit[B: TypeTag](values: B*): Read.Literal[B] = Literal(values.toSeq)
   }
 
   sealed trait Ordering[+A]
@@ -447,11 +446,11 @@ trait Sql {
       def isNumeric: IsNumeric[A] = implicitly[IsNumeric[A]]
     }
 
-    case object AndBool        extends BinaryOp[Boolean]
-    case object OrBool         extends BinaryOp[Boolean]
+    case object AndBool extends BinaryOp[Boolean]
+    case object OrBool  extends BinaryOp[Boolean]
 
-    case object AndBit        extends BinaryOp[Integer]
-    case object OrBit         extends BinaryOp[Integer]
+    case object AndBit extends BinaryOp[Integer]
+    case object OrBit  extends BinaryOp[Integer]
 
   }
   sealed trait RelationalOp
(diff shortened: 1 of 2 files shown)


unclebob418:
@kitlangton it was fun pairing with you! One issue that we haven't yet solved is how to deal with a union of two (or more) single column selects which would be valid (assuming the types align, will have to check that's enforced). I think you've covered everything else in your rather comprehensive writeup!

kitlangton:
Willing to modify future verbosity levels—just wanted to err on the side of TMI for this first PR 😉

unclebob418:
@kitlangton that was a compliment - I think detail is a good thing, especially since we tried a couple of different approaches

jdegoes:
@kitlangton @unclebob418 Have you explored using `F` ("features") for this? 

Currently only `Select` has `F` but we could extend its siblings to support that to surface deep features up into a `Read`. Then introduce an F-based type class that is derived only for suitable shapes.

Might be worth exploring a bit...

kitlangton:
Did some more exploration of why the simple version of the implicit conversion didn't fire, and I came up with solution that works. However, it's also a bit odd to me why this works and the other doesn't. Something to do with how the compiler prioritizes implicit vs. type unification I suppose. 

Here's a minimal reproduction: https://scastie.scala-lang.org/kG7F3FUBR0uTkSpz7PgbhQ

Issue #28, opened 2020-05-24
Support selecting columns using existing column names


unclebob418:
Currently we need to specify the column names using `as`, but some columns already have names defined e.g. for source columns. which were specified when we defined the table. These names should be used as default when "as" is not specified. Perhaps we need a new "Named" `Feature`?

```scala
object Example {
  new Sql { self =>
    import self.ColumnSet._

    val userTable = (uuid("user_id") ++ localDate("dob") ++ string("first_name") ++ string("last_name")).table("users")

    val userId :*: dob :*: fName :*: lName :*: _ = userTable.columns

    //val basicSelect = select { fName ++ lName } from userTable //todo this should compile using column names defined in the table

    // fName and lName already have column names, shouldn't have to do this
    val basicSelectWithAliases = (select {
      (fName as "first_name") ++ (lName as "last_name")
    } from userTable)
  }
}
```

jgoday:
Could the use of implicit conversions be a valid solution ?
I mean, we can implicit convert an Expression to a Selection and then use the existing ++ operator.

```scala
def columnName[F, A, B](expr: Expr[F, A, B]): Option[String] =
    expr match {
      case Expr.Source(_, c) => Some(c.name)
      case _ => None
    }

implicit def expToSelection[F, A, B](expr: Expr[F, A, B]) =
  Selection.computedOption(expr, columnName(expr))
```

and then we can use it like the previous example
```scala
val basicSelect = select { fName ++ lName } from userTable //todo this should compile using column names defined in the table

// fName and lName already have column names, shouldn't have to do this
val basicSelectWithAliases = (select {
  (fName as "first_name") ++ (lName as "last_name")
} from userTable)

...

test("basic select") {
   assert(basicSelect)(equalTo(basicSelectWithAliases))
}

```

Representation of basicSelect is in fact =
```scala
Select(Selection(
    Cons(Computed(Source(users,Column(first_name)),Some(first_name)),
    Cons(Computed(Source(users,Column(last_name)),Some(last_name)),Empty))),
     zio.sql.Sql$ColumnSet$Cons$$anon$1@782715e4,Literal(true),List(),List(),None,None)
```

unclebob418:
I think it will have to be an implicit conversion, but I think it needs to be a little more complicated to support things other that sources that have been named e.g. below we still have to specify a name for Arbitrary(age), and count isn't a source.

```scala
val aggregated =
      (select {
[...]

PR #47, opened 2020-07-17
tidy up examples (#28 fixed)

unclebob418:
#28 has been fixed so we don't need aliases everywhere.

### DIFF
--- a/zio-sql/jvm/src/main/scala/zio/sql/Examples.scala
+++ b/zio-sql/jvm/src/main/scala/zio/sql/Examples.scala
@@ -10,25 +10,16 @@ object Examples {
   import ShopSchema.OrderDetails._
 
   //select first_name, last_name from users
-  //val basicSelect = select { fName ++ lName } from userTable //todo fix issue #28
+  val basicSelect = select { fName ++ lName } from users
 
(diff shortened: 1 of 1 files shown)


zio-sql/jvm/src/main/scala/zio/sql/Examples.scala
```diff
@@ -10,33 +10,24 @@ object Examples {
...
-      (fName as "first_name") ++ (lName as "last_name") //todo fix #28 get rid of aliases
-    }
-      from users)
-      .limit(2)
-      .orderBy(
-        lName //defaults to ascending order, do not need to specify .asc
-        ,
-        fName.desc //desc must be specified if you want descending order
-      )
+    select { fName ++ lName } from users orderBy (lName, fName.desc) limit 2
```
lemastero, line 22: I wonder if it is possible to keep orderBy part?
```scala
.orderBy(	
        lName //defaults to ascending order, do not need to specify .asc	
        ,	
        fName.desc //desc must be specified if you want descending order	
      )
```

jdegoes:
@unclebob418 Great work on this, thank you!

PR #48, opened 2020-07-25
Add sbt-tpolecat and move examples to sub-project

lemastero:
Changes:
* move Examples into separate sub-project - similar approach is in zio, zio-nio,
* scalafmtAll
* add plugin `sbt-tpolecat` as in `zio-nio`


softinio:
@lemastero @mijicd Most projects aren't using `sbt-tpolecat` I am fine with it, but if we want to start standardizing on it more widely shall we add it to our seed?

mijicd:
@softinio agreed.

lemastero:
Closing this guy. I wanted to improve this project using some principles I try to follow when writing code:
* turn on all static code analysis from the compiler (via sbt-topolecat plugin)
* put code in proper packages
* separate real code from samples

But maybe I am too opinionated :)

mijicd:
@lemastero these makes perfect sense actually. What @softinio pointed above is that it would be nice to have them in our library giter8 template, so that the other projects are started with proper setup :).

PR #49, opened 2020-07-27
add unary operators

### LINKED ISSUE #4: Add all unary operators supported by ANSI-92
We should support the full set of unary operators, like `NOT`, `-`, `IS NULL`, etc., which are defined by ANSI-92 and / or supported by Postgres, MySQL, SQL Server, and Oracle.

These would be added as `UnaryOp` subtypes, with new methods defined in `Expr`.

If anything is unclear, please ask for more details!

lemastero:
Fix #4 add unary operators


zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -523,6 +537,21 @@ trait Sql {
     )(implicit ev: B <:< Integer): Expr[F :||: F2, A1, Integer] =
       Expr.Binary(self.widen[Integer], that, BinaryOp.OrBit)
 
+    def ~[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
```
jdegoes, line 540: Unary operators can be written like this:

```scala
def unary_~ 
```

In order to be allowed to use them as unary operators in Scala.

Same for the other unary operators below.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -523,6 +537,21 @@ trait Sql {
...
+      Expr.Unary(self.widen[Integer], UnaryOp.NotBit)
+
+    def -[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
+      Expr.Unary(self.widen[Integer], UnaryOp.NotNum)
+
+    def not[A1 <: A]()(implicit ev: B <:< Boolean): Expr.Unary[F, A1, Boolean] =
+      Expr.Unary(self.widen[Boolean], UnaryOp.NotBool)
+
+    def isNull[A1 <: A](): Expr.Unary[F, A1, Any] =
```
jdegoes, line 549: There is no need for empty parentheses. In Scala, empty parens = side-effect. But ZIO SQL is a pure functional library and has no side effects.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -523,6 +537,21 @@ trait Sql {
     )(implicit ev: B <:< Integer): Expr[F :||: F2, A1, Integer] =
       Expr.Binary(self.widen[Integer], that, BinaryOp.OrBit)
 
+    def ~[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
+      Expr.Unary(self.widen[Integer], UnaryOp.NotBit)
+
+    def -[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
```
jdegoes, line 543: No need for empty parens here.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -523,6 +537,21 @@ trait Sql {
...
+      Expr.Unary(self.widen[Integer], UnaryOp.NotNum)
+
+    def not[A1 <: A]()(implicit ev: B <:< Boolean): Expr.Unary[F, A1, Boolean] =
+      Expr.Unary(self.widen[Boolean], UnaryOp.NotBool)
+
+    def isNull[A1 <: A](): Expr.Unary[F, A1, Any] =
+      Expr.Unary(self.widen[Any], UnaryOp.IsNull)
+
+    def isNotNull[A1 <: A](): Expr.Unary[F, A1, Any] =
```
jdegoes, line 552: No need for empty parens here.

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -427,6 +427,20 @@ trait Sql {
     }
   }
 
+  sealed trait UnaryOp[A]
+
+  object UnaryOp {
+    sealed case class Not[A: IsNumeric]() extends UnaryOp[A] {
```
jdegoes, line 433: I think you mean `Negate`, right?

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: Thank you. I did not know. Changed using `unary_`

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: Very good point. Removed ()

zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -523,6 +537,21 @@ trait Sql {
     )(implicit ev: B <:< Integer): Expr[F :||: F2, A1, Integer] =
       Expr.Binary(self.widen[Integer], that, BinaryOp.OrBit)
 
+    def ~[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
+      Expr.Unary(self.widen[Integer], UnaryOp.NotBit)
+
+    def -[A1 <: A]()(implicit ev: B <:< Integer): Expr.Unary[F, A1, Integer] =
+      Expr.Unary(self.widen[Integer], UnaryOp.NotNum)
+
+    def not[A1 <: A]()(implicit ev: B <:< Boolean): Expr.Unary[F, A1, Boolean] =
```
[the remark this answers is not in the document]
lemastero: Removed ()

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: Removed ()

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: Removed ()

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: I like that name. Thank you :)

lemastero:
I have problems implementing tests for other unary ops. I wrote them based on binary ops. Possibly there is some incompatibility between column type and operators type. I am investigating this ATM.

zio-sql/jvm/src/test/scala/zio/sql/UnaryOpSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+        string("name") ++
+        int("amount") ++
+        boolean("deleted")
+      ).table("product")
+
+    val id :*: lastUpdated :*: name :*: amount :*: deleted :*: _ = productTable.columns
+
+    val selectAll = select { id ++ lastUpdated ++ amount ++ deleted } from productTable
+
+    val selectDeteltedNotTrue = selectAll.where( deleted.isNotTrue )
```
ivanopagano, line 22: typo ... Detelted

lemastero: Renamed. In fact, I renamed all queries to `query`

(In the different contexts - in the business app - that, most likely, would be an antipattern. But here the name do not give us anything. Actually it forces you to focus on irrelevant business logic. Important is how this logic is implemented. That is only relevant to open source and internal libraries.)

lemastero:
I was not able to write tests for bitwise operators [BitwiseOpSpec](https://github.com/zio/zio-sql/pull/49/files#diff-0c6e2ac99dae4ffb16cda185b1b0f14e). So I change them slightly (both binary and unary): https://github.com/zio/zio-sql/pull/49/commits/ba818bcbd79b18d6aeac682ab7ea9d17cc6cd889
cc: @jczuchnowski

Now they support any numeric type rather than just Java Integer.

Another approach would be to change the type supported by them to `Byte` (Integer seems very arbitrary - maybe I lack some important context). But it seems DB's support binary ops on any numeric column (binary string):
[MS SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-operators-transact-sql?view=sql-server-ver15), [MySQL](https://dev.mysql.com/doc/refman/8.0/en/bit-functions.html), [PostgreSQL](https://www.postgresql.org/docs/current/functions-bitstring.html) for Oracle ... I am not sure if [they support binary operations at all via operators only as functions](https://docs.oracle.com/cd/B28359_01/server.111/b28286/functions014.htm#SQLRF00612).

jczuchnowski:
@lemastero PostreSQL documentation (https://www.postgresql.org/docs/12/functions-math.html) says that bitwise operators work only for integral types. So our numeric doesn't fit here.

lemastero:
@jczuchnowski  How about this:

```scala
sealed trait IsIntegral[A] {
  def typeTag: TypeTag[A]
}

object IsIntegral {

  abstract class AbstractIsIntegral[A: TypeTag] extends IsIntegral[A] {
    def typeTag = implicitly[TypeTag[A]]
  }
  implicit case object TByteIsIntegral       extends AbstractIsIntegral[Byte]
  implicit case object TShortIsIntegral      extends AbstractIsIntegral[Short]
  implicit case object TIntIsIntegral        extends AbstractIsIntegral[Int]
  implicit case object TLongIsIntegral       extends AbstractIsIntegral[Long]
}
```
?

jdegoes:
@lemastero Excellent work!

Issue #38, opened 2020-05-30
all non-aggregated columns must be included in `groupBy` - remove `Arbitrary`?


robmwalsh:
`groupBy` should ensure all non-aggregated columns in the `select` are included in group by (but are not required to be in the same order)

I think perhaps `Arbitrary` should be removed as it makes the columns into `AggregationDef`s even though they are not aggregated.

```scala
val orderValue = (select {
    (Arbitrary(fkUserId) as "id") ++ (Arbitrary(orderId) as "id") ++ (Sum(quantity * unitPrice) as "quantity")
  } from (orders join orderDetails).on(fkOrderId === orderId)) groupBy (fkUserId/* , orderId */) //shouldn't compile without orderId being uncommented
```

PR #51, opened 2020-07-31
add having clause

### LINKED ISSUE #50: Add HAVING
Add support for `having` https://www.w3schools.com/sql/sql_having.asp

In most cases when I use `group by` there is need to add extra conditions using having clause.

lemastero:
Fix #50 add support for `having` clause

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -290,6 +290,7 @@ trait Sql {
       table: Table.Aux[A],
       whereExpr: Expr[_, A, Boolean],
       groupBy: List[Expr[_, A, Any]],
+      havingExpr: Expr[_, A, Boolean] = true,
       orderBy: List[Ordering[Expr[_, A, Any]]] = Nil,
       offset: Option[Long] = None,
       limit: Option[Long] = None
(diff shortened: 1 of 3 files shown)


zio-sql/jvm/src/main/scala/Sql.scala
```diff
@@ -311,6 +312,11 @@ trait Sql {
         val _ = ev
         copy(groupBy = groupBy ++ (key :: keys.toList))
       }
+
+      def having(havingExpr2: Expr[_, A, Boolean])(
```
jdegoes, line 316: Don't we need to assert on `F` (the feature) that everything is aggregated inside it, to conform to restrictions of having clause?

zio-sql/jvm/src/test/scala/zio/sql/GroupByHavinSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+  val productTable = (
+    string("id") ++
+      string("name") ++
+      int("amount") ++
+      double("price")
+  ).table("product")
+
+  val id :*: name :*: amount :*: price :*: _ = productTable.columns
+
+  val orderValue =
```
jdegoes, line 32: Love the example!

zio-sql/jvm/src/main/scala/Sql.scala
lemastero: Related issue: https://github.com/zio/zio-sql/issues/38

PR #53, opened 2020-08-14
Add features to read

robmwalsh:
@jdegoes @kitlangton I had a go at changing Read to  "extend [F (features) to] its siblings to support that to surface deep featusqres up into a Read." as suggested by @jdegoes in #46 which seemed to work fine (that's the first commit here). 

I explored bringing the still-brilliantly named `Inable` to the party but I'm getting an error "Examples.scala:29:57: a type was inferred to be `Any`; this may indicate a programming error. [29:57]" VS Code knows the types (IJ of course doesn't stand a chance here) so I think this may be a compiler error (perhaps related to https://github.com/scala/bug/issues/11798 ?)

Regardless of solving this issue, I don't think we're going to have much luck with the current encoding as there are too many ways things get stuffed into a `Features.Union`. Fundamentally I think we're mixing abstractions together with `Union`. It's used to represent the features of multiple columns but also used for all the variations of `FunctionCall` which `Union`s the `Features` of all the parameters, so Substring on a single column is going to have several levels of `Union` for the one column.

Would it be a reasonable approach to have a few different type of Unions (e.g. one for adding columns, one for adding rows i.e. SQL's union, one for the various function parameters) to avoid this ambiguity?

### DIFF
--- a/zio-sql/jvm/src/main/scala/Sql.scala
+++ b/zio-sql/jvm/src/main/scala/Sql.scala
@@ -372,7 +372,7 @@ trait Sql {
 
     def ++[F2, A1 <: A, C <: SelectionSet[A1]](
       that: Selection[F2, A1, C]
-    ): Selection[F :||: F2, A1, self.value.Append[A1, C]] =
+    ): Selection[F :><: F2, A1, self.value.Append[A1, C]] =
       Selection(self.value ++ that.value)
 
     def columns[A1 <: A]: value.SelectionsRepr[A1, SelectionType] = value.selections[A1, SelectionType]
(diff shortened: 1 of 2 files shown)


jdegoes:
@robmwalsh I do think maybe Union is poorly defined. It means to say "something beneath this level has these features", but probably we need notion of product versus coproduct: everything beneath here has this feature, or something beneath here has this feature. That would at least be a start at introducing more precision.

jdegoes:
After thinking about this in some depth, I believe that we can reflect the type of a term in the phantom type parameter itself. e.g. `Expr.Int` would have `Expr.Int` in the type parameter. This means there is a 1-to-1 correspondence between how an expression is constructed and the phantom type parameter. This 'lossless' encoding may complicated some things but would ensure the phantom type is well-defined over all expressions.

The technique could be pushed further up beyond Expr in a similar way.

jdegoes:
Will close this for now because it has bit-rotted.

Issue #66, opened 2020-09-08
ExecuteBuilder in ReadExecutor only supports arities 1-2


jdegoes:
Support for arities 1-20 should be easy to add by following the pattern established in `ExecuteBuilder`.

Please request clarification if anything is unclear.

phderome:
fyi: The quill project goes through some pains to have names of Scala variables in the lambda (`to` in ZIO-SQL) being tied when possible to column aliases with the queries. This is a valuable feature.

jacobBaumbach:
I would like to give this one a go

PR #78, opened 2020-09-10
issue66

jacobBaumbach:
This PR solves #66 

I added support for arities 3-20 and ran fmt on the file.

One minor issue I had was assigning a name to value of type `F`.  Since `f` is already being used for the function, I decided to go with the name `fArg` for the value of type `F`.  If there is a different name I should use, please let me know.


jdegoes:
@jacobBaumbach Awesome, thank you!

Issue #2, opened 2020-05-19
Split up giant Sql.scala file into multiple files


jdegoes:
Currently, Sql.scala defines a module `trait Sql`, in which everything is placed. This means all data types and functions are located in a single file.

To make maintenance easier and improve organization, we should split this file into many files. However, because we are using the module pattern, this means we must not split just the file, but also split the `trait Sql` module into multiple submodules, which will depend on each other (initially using `trait XYZ extends ZYX`, probably).

Looking at the way the file is organized now, it looks like we could pull out the following modules:

 - `trait NewtypesModule` / `newtypes.scala`, which would contain things like ColumnName, TableName, FunctionName, etc. Simple type aliases or case class wrappers, which have no dependencies and just wrap other types.
 - `trait TypeTagModule` / `typetag.scala`, which would contain `TypeTag` and companion object, with the extension alias. Possibly, it could contain `IsNumeric` and other 
 - `trait FeaturesModule` / `features.scala` which would contain `Features` and associated stuff
 - `trait ExprModule` / `expr.scala` (extends `NewtypeModule`, `TypeTagModule`), which would contain `Expr`, together with `FunctionDef` and `AggregationDef` and their helpers
 - `trait TableModule` / `table.scala` (extends `ExprModule`), which could contain `ColumnSchema`, `ColumnSet`, `Column`, `JoinType`, `Table`.
 - `trait SelectModule` / `select.scala`, which contains all Select related classes and helpers.
 - `trait DeleteModule` / `delete.scala`, which contains all Delete related classes and helpers.
 - `trait UpdateModule` / `update.scala`, which contains all Update related classes and helpers.

It would be nice to find a way to pull out the `FunctionDef` and `AggregationDef` listings into their  own separate files / modules, without circular dependencies, as these will contain long lists of very simple definitions. That will require a bit of work.

FrancisToth:
@jdegoes :  We (@calvinlfer and myself) could work on this if you don't mind.

jdegoes:
Superb!

jczuchnowski:
@FrancisToth @calvinlfer are you still working on this? I suspect you might have quite a lot of conflicts after recent changes.

jczuchnowski:
Actually asking as I can give you a hand over the weekend ;)

robmwalsh:
@jczuchnowski happy to pair with you on this one if @calvinlfer and @FrancisToth are unavailable. Last time I checked (on discord) they hadn't started on it.

[...]

PR #81, opened 2020-09-17
breakup Sql file into modules

jczuchnowski:
Addresses https://github.com/zio/zio-sql/issues/2

Result of collaborative work with @robmwalsh.


jdegoes (approved):
Awesome work and much needed!

PR #100, opened 2020-09-24
[WIP] Sql server rendering

### LINKED ISSUE #38: all non-aggregated columns must be included in `groupBy` - remove `Arbitrary`?

robmwalsh:
I've got SQL server rendering most of the way there (select works*, insert/update/delete  should be easy enough from here) + a bit of housekeeping on the build config.

[1c037cf](https://github.com/zio/zio-sql/commit/1c037cf8fbd30db1b7a15f65d11ded6764ca141f) was able to render the selects from `Examples` in a way that were able to be executed unmodified (apart from removing "arbitrary" functions which don't exist in TSQL) against a sql server instance with the test data.

~~Unfortunately, deleting `Renderable` (which was no longer used) has caused an error "The outer reference in this type test cannot be checked at run time" (line 186 of SqlServerModule). I think this may relate to https://github.com/scalameta/scalameta/issues/900 but I'm not sure. I don't really understand what the suggested workaround is trying to fix. **This has got me stumped, any help on this would be appreciated!**~~ "solved" with the help of @kitlangton

Still to do
1. render constants/variables properly including sql injection protection
1. insert/update/delete
1. figure out which bits can be reused across dialects & extract to SQL module
1. fix #38 which actually needs to be fixed now because sql server doesn't support Arbitrary


robmwalsh:
cc @jdegoes / @jczuchnowski

jdegoes:
This is awesome work.

For (4), please just use `MIN` wherever you see `ARBITRARY`. Also, I'd ignore (3). Once we have a lot of duplication across different SQL Dialects, we can then worry about how to factor it out. Right now we don't really know what that's going to look like.

robmwalsh:
Thanks @jdegoes

for escaping/proper rendering of values I'll also need to solve the typeTypeTagOf issue for it to actually work but I think that's the only real blocker at this point.

Do you have any idea why Table needs to extend `Insanity` (or more specifically any trait with some def) in order to be able to pattern match against it? That's a bit of ugliness I'd ideally like to get rid of.

jdegoes:
@robmwalsh I know that @jczuchnowski has implemented `typeTagOf`, so maybe he can donate that here. 😄 

No idea, but I suspect bug in Scalac.

robmwalsh:
@jczuchnowski any chance of a copy of your typeTagOf implementation please?

jczuchnowski:
@robmwalsh I don't have a full implementation yet, but I'll post what I have and maybe @jdegoes will help us move forward ;)

I would also leave out 2. from this PR. This PR is already very useful. I'm working on integrating Testcontainers with some basic integration test and I could use what we have here already.

robmwalsh:
@jdegoes @jczuchnowski shall we just merge this now and I can continue rendering work separately?

jczuchnowski:
Yeah, I'd merge it.

jdegoes (approved):

PR #117, opened 2020-10-26
End to end test for PostgreSQL with Testcontainers

jczuchnowski:
This PR introduces Testcontainers for integration testing. That also proves we're actually able to do an SQL query (yay! :fireworks:  )

I had to change the implementation of ReadExecutor, because it was returning only the first row.

There's a to do after this is merged (possibly should be turned into issues):
- generalize Testcontainers integration to be usable for all database modules
- make ZIO Test integrate better with ZIO SQL
- complete the implementatoin of `Expr.typeTagOf` (I made it just to handle the basic queries)
- fix ReadExecutor to handle aggregation functions (I left a commented integration test that shows a problem)


postgres/src/test/scala/zio/sql/TestContainers.scala
```diff
@@ -0,0 +1,33 @@
...
+import zio.blocking.{ effectBlocking, Blocking }
+
+object TestContainer {
+
+  def container[C <: SingleContainer[_]: Tag](c: C): ZLayer[Blocking, Nothing, Has[C]] =
+    ZManaged.make {
+      effectBlocking {
+        c.start()
+        c
+      }.orDie
```
robmwalsh, line 15: Might be a good idea to handle the error so we get nicer test failures if the container doesn't start. I don't like seeing fiber failures in test results because they take up so much space in the console. I think the probability of a container not starting is high enough to handle it.

postgres/src/test/scala/zio/sql/postgresql/PostgresModuleTest.scala
```diff
@@ -0,0 +1,130 @@
...
+          Customer(row._1, row._2, row._3, row._4)
+        }
+
+      val assertion = for {
+        r <- testResult.runCollect
+      } yield assert(r)(hasSameElementsDistinct(expected))
+
+      assertion.provideCustomLayer(executorLayer)
+    },
+    // testM("Can count rows") {
```
robmwalsh, line 71: Can you please either delete or leave a comment  for what needs to be fixed to this to be uncommented? I'm guessing this relates to your comment about aggregations.

postgres/src/test/scala/zio/sql/postgresql/PostgresModuleTest.scala
```diff
@@ -0,0 +1,130 @@
...
+    //   } yield assert(r.head)(equalTo(expected))
+    // },
+    testM("Can select from joined tables (inner join)") {
+      val query = select(fName ++ lName ++ orderDate) from (customers join orders).on(fkCustomerId === customerId)
+
+      println(renderRead(query))
+
+      case class Row(firstName: String, lastName: String, orderDate: LocalDate)
+
+      val expected = Seq(
```
robmwalsh, line 89: I'm not sure if we can always expect these in a particular order - perhaps a set would be better for now? Ideally we should probably use a bag as there could be duplicate results, but I don't think we get one of those with scala so would have to add it.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -1,5 +1,205 @@
...
+    }
+
+    def buildReadString[A <: SelectionSet[_]](read: self.Read[_]): Unit =
+      read match {
+        //todo offset (needs orderBy, must use fetch _instead_ of top)
+        case Read.Select(selection, table, whereExpr, groupBy, havingExpr, orderBy, offset, limit) =>
+          builder.append("select ")
+          limit match {
+            case Some(limit) =>
+              builder.append("top ").append(limit).append(" ")
```
robmwalsh, line 78: I believe top is a sql server only special. This whole function should be reviewed to see if there's any other sql server/postgres differences. for postgres I believe for postgres this needs to be "limit x" at the end of the statement. the offset issue from the comment should be a non-issue for postgres.

postgres/src/test/scala/zio/sql/postgresql/PostgresModuleTest.scala
```diff
@@ -0,0 +1,130 @@
...
+          Customer(
+            UUID.fromString("636ae137-5b1a-4c8c-b11f-c47c624d9cdc"),
+            "Jose",
+            "Wiggins",
+            LocalDate.parse("1987-03-23")
+          )
+        )
+
+      val testResult = new ExecuteBuilder(query)
+        .to[UUID, String, String, LocalDate, Customer] { case row =>
```
robmwalsh, line 61: not for fixing with this PR but I think we need to come up with a better way of extracting test results. We shouldn't have to specify the types as this is encoded in the type tags... just need to figure out how to get them!

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -1,5 +1,205 @@
 package zio.sql.postgresql
 
-import zio.sql.{ Jdbc, Sql }
+import zio.sql.Jdbc
 
-trait PostgresModule extends Sql with Jdbc {}
+trait PostgresModule extends Jdbc { self =>
+
+  override def renderRead(read: self.Read[_]): String = {
```
robmwalsh, line 7: I don't really like having so much code duplicated across the different modules, but I'm also unsure how we might fix this given the path dependent types

robmwalsh (changes requested):
Nice work! Does this mean we can delete the docker compose file or is that still useful? There are a few things to fix up and some general comments but looks pretty good to me overall

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
jczuchnowski: It might look like a code duplication now, but these renderers will diverge more and more over time I think. Too many differences between DBs even at this level.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
jczuchnowski: Sorry, missed that. And yeah, that's actually one of the reasons we need to have these renderers separate.

postgres/src/test/scala/zio/sql/postgresql/PostgresModuleTest.scala
jczuchnowski: Yes, exactly. And sure, I'll add a comment.

jczuchnowski:
@robmwalsh thanks for the review. I'll make the changes this evening. As to removing the scripts - I'm a little hesitant for now, but eventually I think they can be removed. Maybe let's add Testcontainers for other DBs so that we're sure we got it right, and then we can remove them.

jdegoes (commented):
Good to merge when feedback addressed. Excited for this one! 🎉 💪

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
robmwalsh: Very true. At least they all start with "select" :)

PR #140, opened 2020-11-10
Fix isNotTrue PropertyOp

jczuchnowski:


jdegoes (dismissed):
Some targets are failing, but good to merge when updated to master and the build is passing!

Issue #215, opened 2020-11-15
Add 'to_timestamp' function to PostgresModule


jczuchnowski:
Add support for function `to_timestamp` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-datetime.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

brbrown25:
I'll pick this one up

robmwalsh:
Fixed in #263

PR #263, opened 2020-11-21
feat(postgres to_timestamp): Implementing Postgres to_timestamp funct…

brbrown25:
…ion. #215

### DIFF
--- a/jdbc/src/main/scala/zio/sql/jdbc.scala
+++ b/jdbc/src/main/scala/zio/sql/jdbc.scala
@@ -1,9 +1,8 @@
 package zio.sql
 
 import java.sql._
-
 import java.io.IOException
-
+import java.time.{ ZoneId, ZoneOffset }
 import zio.{ Chunk, Has, IO, Managed, ZIO, ZLayer, ZManaged }
 import zio.blocking.Blocking
 import zio.stream.{ Stream, ZStream }
@@ -193,7 +192,14 @@ trait Jdbc extends zio.sql.Sql {
           tryDecode[java.util.UUID](
             java.util.UUID.fromString(column.fold(resultSet.getString(_), resultSet.getString(_)))
           )
-        case TZonedDateTime      => ???
+        case TZonedDateTime      =>
+          tryDecode[java.time.ZonedDateTime](
+            java.time.ZonedDateTime
+              .ofInstant(
+                column.fold(resultSet.getTimestamp(_), resultSet.getTimestamp(_)).toInstant,
+                ZoneId.of(ZoneOffset.UTC.getId)
+              )
+          )
         case TDialectSpecific(_) => ???
         case t @ Nullable()      => extractColumn(column, resultSet, t.typeTag, false).map(Option(_))
       }
--- a/postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
+++ b/postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
@@ -1,7 +1,6 @@
 package zio.sql.postgresql
 
-import java.time.LocalDate
-
+import java.time.{ LocalDate, ZonedDateTime }
 import zio.sql.Jdbc
 
 /**
@@ -30,6 +29,7 @@ trait PostgresModule extends Jdbc { self =>
     val Degrees     = FunctionDef[Double, Double](FunctionName("degrees"))
     val Div         = FunctionDef[(Double, Double), Double](FunctionName("div"))
     val Factorial   = FunctionDef[Int, Int](FunctionName("factorial"))
+    val ToTimestamp = FunctionDef[Long, ZonedDateTime](FunctionName("to_timestamp"))
   }
 
   override def renderRead(read: self.Read[_]): String = {
--- a/postgres/src/test/resources/shop_schema.sql
+++ b/postgres/src/test/resources/shop_schema.sql
@@ -37,6 +37,12 @@ create table order_details
     unit_price money not null
 );
 
+create table timestamp_test
+(
+    timestamp_id uuid not null,
+    created_timestamp_string varchar not null,
+    created_timestamp timestamp with time zone default now()
+);
 
 insert into customers
     (id, first_name, last_name, verified, dob)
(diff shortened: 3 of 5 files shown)


jdbc/src/main/scala/zio/sql/jdbc.scala
```diff
@@ -193,7 +192,14 @@ trait Jdbc extends zio.sql.Sql {
           tryDecode[java.util.UUID](
             java.util.UUID.fromString(column.fold(resultSet.getString(_), resultSet.getString(_)))
           )
-        case TZonedDateTime      => ???
+        case TZonedDateTime      =>
+          tryDecode[java.time.ZonedDateTime](
+            java.time.ZonedDateTime
+              .ofInstant(
+                column.fold(resultSet.getTimestamp(_), resultSet.getTimestamp(_)).toInstant,
+                ZoneId.of(ZoneOffset.UTC.getId)
```
brbrown25, line 200: I'm not sure how we'd like to handle this, having it default to utc seems problematic, but I'm not sure how we should go from java.sql.Timestamp -> java.time.ZonedDateTime.

robmwalsh: looking at the javadoc of getTimestamp, I'm pretty sure that it's converting to UTC, so I believe this is correct. It'd probably be a good idea to add a round trip test to make sure (e.g. insert a timestamp from some other time zone, grab it back out, then compare to ensure that time is the same (accounting for the difference in time zones)

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -1,13 +1,15 @@
...
 import zio.sql.Jdbc
 
 /**
  */
 trait PostgresModule extends Jdbc { self =>
 
   object PostgresFunctionDef {
-    val Sind = FunctionDef[Double, Double](FunctionName("sind"))
+    val Sind        = FunctionDef[Double, Double](FunctionName("sind"))
+    val ToTimestamp = FunctionDef[Double, ZonedDateTime](FunctionName("to_timestamp"))
```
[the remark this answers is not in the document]
brbrown25: That makes sense I was mostly copying from Postgres docs, I'll make that change!

jdbc/src/main/scala/zio/sql/jdbc.scala
brbrown25: Yeah that's a good idea, I'll add that in!

brbrown25:
@robmwalsh I've added in some additional tests per your suggestion and I've brought in latest master. Ready for rereview.

postgres/src/test/resources/shop_schema.sql
```diff
@@ -189,4 +195,12 @@ values
     ('852E2DC9-4EC3-4225-A6F7-4F42F8FF728E', 'D5137D3A-894A-4109-9986-E982541B434F', 1, 45.45),
     ('D6D8DDDC-4B0B-4D74-8EDC-A54E9B7F35F7', 'D5137D3A-894A-4109-9986-E982541B434F', 2, 50.00),
     ('2C3FC180-D0DF-4D7B-A271-E6CCD2440393', 'D5137D3A-894A-4109-9986-E982541B434F', 2, 50.00),
-    ('5883CB62-D792-4EE3-ACBC-FE85B6BAA998', 'D5137D3A-894A-4109-9986-E982541B434F', 1, 55.00);
\ No newline at end of file
+    ('5883CB62-D792-4EE3-ACBC-FE85B6BAA998', 'D5137D3A-894A-4109-9986-E982541B434F', 1, 55.00);
+
+insert into timestamp_test
```
robmwalsh, line 200: I actually had in mind inserting rows using a zio-sql query, but doesn't look like we support that at the moment. Can you please add the created_timestamp and created_timestamp_string columns to the customer table instead of creating this new one (which looks like the approach you were taking before) and add a note to remove the created_timestamp_string column once #320 is done?

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -523,6 +522,34 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+    testM("to_timestamp") {
+      import this.TimestampTests._
+
+      val query      = select(ToTimestamp(1284352323L)) from customers
+      val expected   = ZonedDateTime.of(2010, 9, 13, 4, 32, 3, 0, ZoneId.of(ZoneOffset.UTC.getId))
+      val testResult = execute(query).to[ZonedDateTime, ZonedDateTime](identity)
+
+      val expectedRoundTripTimestamp = ZonedDateTime.of(2020, 11, 21, 19, 10, 25, 0, ZoneId.of(ZoneOffset.UTC.getId))
+      val roundTripQuery             =
+        select(createdString ++ createdTimestamp) from timestampTests
```
robmwalsh, line 536: `select(createdString ++ createdTimestamp) from customers` - see previous comment

postgres/src/test/scala/zio/sql/postgresql/ShopSchema.scala
```diff
@@ -7,10 +7,12 @@ trait ShopSchema extends Jdbc { self =>
 
   object Customers     {
     val customers =
-      (uuid("id") ++ localDate("dob") ++ string("first_name") ++ string("last_name") ++ boolean("verified"))
+      (uuid("id") ++ localDate("dob") ++ string("first_name") ++ string("last_name") ++ boolean(
+        "verified"
+      ) ++ zonedDateTime("created_timestamp"))
```
robmwalsh, line 12: I think the approach you were starting here looks better than adding a separate table

robmwalsh (changes requested):
awesome work - just a couple of fairly minor changes and this should be good to go

postgres/src/test/resources/shop_schema.sql
brbrown25: > I actually had in mind inserting rows using a zio-sql query, 
Yes I can do that. my original plan was to do that as well. I'll make a note and your requested change!

- [ ] using existing customers table
- [ ] add note

jczuchnowski:
So is this ready? It looks ready...

brbrown25:
> So is this ready? It looks ready...

Yep this is ready now!

jczuchnowski:
Awesome progress!

jczuchnowski (approved):

robmwalsh:
@brbrown25 thanks!

Issue #183, opened 2020-11-14
Add 'concat_ws' function to PostgresModule


jczuchnowski:
Add support for function `concat_ws` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-string.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

pjazdzewski1990:
I would like to work on this.

PR #267, opened 2020-11-21
issue/183 concat_ws 

pjazdzewski1990:
For issue https://github.com/zio/zio-sql/issues/183


jczuchnowski:
I think this is good enough. We need to figure out how to handle vararg functions but it can be done some other time.

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -108,6 +108,7 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
 
     def typeTagOf[A](expr: Expr[_, _, A]): TypeTag[A] = expr.asInstanceOf[InvariantExpr[_, _, A]].typeTag
 
+    implicit def literal(str: String): Expr[Features.Literal, Any, String] = Expr.Literal(s"'${str}'")
```
jczuchnowski, line 111: Please remove this. We need to deal with at the `renderRead` level.

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -279,6 +280,9 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
     val Ascii       = FunctionDef[String, Int](FunctionName("ascii"))
     val CharLength  = FunctionDef[String, Int](FunctionName("character_length"))
     val Concat      = FunctionDef[(String, String), String](FunctionName("concat"))
+    val ConcatWs2   = FunctionDef[(String, String), String](FunctionName("concat_ws"))
```
jczuchnowski, line 283: So, `concat_ws` is not available at least for Oracle, so we need to do it in the db-specific modules. For now it's fine to do it in PostgresModule, although SQL Server also supports this function.

jczuchnowski:
@pjazdzewski1990 couple of problems after all

core/jvm/src/main/scala/zio/sql/expr.scala
jczuchnowski: You can add single quotes to the parameters in tests. We'll fix it once we fix `renderRead`.

pjazdzewski1990: Done

core/jvm/src/main/scala/zio/sql/expr.scala
pjazdzewski1990: Ok, in another PR

jczuchnowski (approved):

Issue #199, opened 2020-11-15
Add 'split_part' function to PostgresModule


jczuchnowski:
Add support for function `split_part` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-string.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

xplosunn:
I'll tackle this one.

PR #269, opened 2020-11-21
Add 'split_part' function to PostgresModule

xplosunn:
Solves #199


core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -108,7 +108,11 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
 
     def typeTagOf[A](expr: Expr[_, _, A]): TypeTag[A] = expr.asInstanceOf[InvariantExpr[_, _, A]].typeTag
 
-    implicit def literal[A: TypeTag](a: A): Expr[Features.Literal, Any, A] = Expr.Literal(a)
+    implicit def literal[A](a: A)(implicit typeTag: TypeTag[A]): Expr[Features.Literal, Any, A] =
+      typeTag match {
+        case TypeTag.TString => Expr.Literal(s"'${a.toString}'").asInstanceOf[Expr[Features.Literal, Any, A]]
```
jczuchnowski, line 113: This is how String should be interpreted/rendered, so it a better fit to do this operation in PostgresModule#renderRead

jczuchnowski (dismissed):

jczuchnowski (approved):

Issue #258, opened 2020-11-21
Support support context bounds on `FunctionDef`s


robmwalsh:
There are lots of `FunctionDef`'s that should work for multiple types specified by existing parameter bounds e.g. 
```scala
val Sign = FunctionDef[Double, Int](FunctionName("sign"))
```
it would be good to be able to do something like:
```scala
def Sign[A: IsNumeric] = FunctionDef[A, Int](FunctionName("sign")) 
```
This doesn't work because we're using the FunctionDef syntax class - the type of the argument passed to the `apply` method on the returned `FunctionCall` is not known when `Sign` is called, meaning we have to specify the type when calling the function and explicity call apply which is not exactly a nice nice UX

```scala
select(Sign[Double].apply(30.0)) from ???
```

this issue was discovered from some work with @saucam & @brbrown25 on other functions

jdegoes:
Yes, we need to fix this. It's a big usability problem.

PR #271, opened 2020-11-21
Context bounds on `FunctionDef`s

xplosunn:
Proposal to address #258

Not sure if this is a good way to go about it.


robmwalsh (changes requested):
Nice concept! I think this should work well with a little more flesh. It'd also be good to have some way to specify an exact type for one of the slots, so you could do something like `val Right = ContextBoundFunctionDef2[Is[String], IsIntegral, String](FunctionName("right"))`

Issue #188, opened 2020-11-14
Add 'lpad' and 'rpad' functions to PostgresModule


jczuchnowski:
Add support for functions `lpad` and `rpad` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-string.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

brbrown25:
I'll pick this up

PR #272, opened 2020-11-21
WIP on adding lpad and rpad. #188

brbrown25:


postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -8,6 +8,8 @@ trait PostgresModule extends Jdbc { self =>
 
   object PostgresFunctionDef {
     val Sind = FunctionDef[Double, Double](FunctionName("sind"))
+    val LPad = FunctionDef[String, Int, String](FunctionName("lpad"))
```
brbrown25, line 11: I get a compile error of 
```
[error] /Users/brown/git/zio/zio-sql/postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala:11:27: wrong number of type parameters for method apply: [A, B](name: PostgresModule.this.FunctionName): PostgresModule.this.FunctionDef[A,B] in object FunctionDef
[error]     val LPad = FunctionDef[String, Int, String](FunctionName("lpad"))
[error]                           ^
[error] /Users/brown/git/zio/zio-sql/postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala:12:27: wrong number of type parameters for method apply: [A, B](name: PostgresModule.this.FunctionName): PostgresModule.this.FunctionDef[A,B] in object FunctionDef
[error]     val RPad = FunctionDef[String, Int, String](FunctionName("rpad"))
[error]                           ^
[error] two errors found
[info] Passed: Total 0, Failed 0
```
Which I'm not quite sure why, as I'd think the apply which calls `Expr.FunctionCall3` would be getting used

brbrown25: ahh because FunctionDef[-A, +B] only takes 2 type params

brbrown25: Yep I figured it out, will have an updated commit soon

angelicarao: Yup, indeed! You can provide the input params for `LPad` as a tuple, so it would look like `[(String, Int), String]`.

brbrown25:
```
[info] - Postgres FunctionDef
[info]   - lpad
[info]     Fiber failed.
[info]     A checked error was not handled.
[info]     org.postgresql.util.PSQLException: ERROR: syntax error at or near "world"
[info]       Position: 19
[info]     No ZIO Trace available.
[info]   - rpad
[info]     Fiber failed.
[info]     A checked error was not handled.
[info]     org.postgresql.util.PSQLException: ERROR: syntax error at or near "world"
[info]       Position: 19
[info]     No ZIO Trace available.
```

jczuchnowski: @brbrown25 it's probably due to a bug we have with rendering String arguments. Try putting them in single quotes like this:

brbrown25: Sweet I'll be updating this either this evening or first thing tomorrow!

brbrown25: @jczuchnowski I've pulled in latest master and applied the escaping change so that the tests pass.

jczuchnowski (approved):

Issue #55, opened 2020-09-03
Support INSERT queries


lemastero:
There is no support for the INSERT clause.

jczuchnowski:
Requires to create `InsertModule` similar to `SelectModule`, `UpdateModule` and `DeleteModule` and declare analogical InsertExecutor in `jdbc.scala` module.
Live implementation of the executor can be done in a different PR.

jotarada:
i can pick this one

jczuchnowski:
Implemented https://github.com/zio/zio-sql/pull/550

PR #279, opened 2020-11-21
new insert module

jotarada:
to solve #55

### DIFF
--- a/core/jvm/src/main/scala/zio/sql/Sql.scala
+++ b/core/jvm/src/main/scala/zio/sql/Sql.scala
@@ -1,6 +1,7 @@
 package zio.sql
 
-trait Sql extends SelectModule with DeleteModule with UpdateModule with ExprModule with TableModule with InsertModule { self =>
+trait Sql extends SelectModule with DeleteModule with UpdateModule with ExprModule with TableModule with InsertModule {
+  self =>
 
   /*
    * (SELECT *, "foo", table.a + table.b AS sum... FROM table WHERE cond) UNION (SELECT ... FROM table)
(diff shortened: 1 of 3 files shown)


jczuchnowski:
We'll have to work on it some more. `SelectionSet` is problematic here after all. It allows to do weird things like `val colsToInsert = age ++ Ascii(name)`.
Also `.values` should take the whole row of values to insert. Otherwise it'll be possible to construct a bad guery.

jotarada:
I think i need help on this.

can `SelectionSet` be replaced with `Selection` ? 

values should be be warped inside a List? the problem will persist because the list can not have all the fields.

jczuchnowski:
@jotarada let me get back to you on that. It's pretty important we get it right.

robmwalsh:
@jotarada I think we should be able use ColumnSet (or we'll have to create something similar) - this would allow us to keep track of the types (and, importantly, the order) of each column specified, ensure that the columns all came from the table specified, and allow us to enforce the types line up in values, so the `values` function could take a vararg of `ValueSet`s (which will need to be created) to have our DSL behave something like. note that the ++ operator in the set of values wouldn't be the same as on ColumnSet
```scala
//should work
insertInto(customers){fName ++ lName ++ dob} values ("Joe" ++ "Bloggs" ++ LocalDate.of(1990, 1, 8), "Matt" ++ "Smith" ++ LocalDate.of(1978, 4, 5))
//shouldn"t work (not enough arguments)
insertInto(customers ){fName ++ lName ++ dob} values ("Joe" ++ "Bloggs", "Matt" ++ "Smith")
//shouldn"t work (order of types doesn"t align)
insertInto(customers ){fName ++ lName ++ dob} values ("Joe" ++ LocalDate.of(1990, 1, 8) ++ "Bloggs", "Matt" ++ LocalDate.of(1990, 1, 8) ++ "Smith")
```
happy to pair with you on this if you'd like

jczuchnowski:
Closing this. In the end it was implemented here https://github.com/zio/zio-sql/pull/550

Issue #150, opened 2020-11-12
Implement renderDelete for PostgrSQL


jczuchnowski:
Following the pattern from `renderRead` in PostgresModule and SqlServerModule implement an PostgresSQL specific `renderDelete`.

kidonm:
I would like to work on this one.

PR #281, opened 2020-11-21
Postgres render delete

kidonm:
fixes: https://github.com/zio/zio-sql/issues/150

This PR adds renderDelete for PostgreSQL 

- implements `DeleteExecutor`
  - makes `execute` return `Int`, the number of deleted rows
- adds `all` combinator to `DeleteBuilder` as `WHERE` clause in `DELETE FROM` is optional
- ~moves rendering functions into `Renderer` to be reusable across `render{Read | Delete | etc...}`~
- adds tests

### DIFF
--- a/core/jvm/src/main/scala/zio/sql/delete.scala
+++ b/core/jvm/src/main/scala/zio/sql/delete.scala
@@ -2,10 +2,10 @@ package zio.sql
 
 trait DeleteModule { self: ExprModule with TableModule =>
 
-  sealed case class DeleteBuilder[A](table: Table.Aux[A]) {
-    def where[F](expr: Expr[F, A, Boolean]): Delete[F, A] = Delete(table, expr)
+  sealed case class DeleteBuilder[F[_], A, B](table: Table.Aux[A]) {
+    def where[F1](expr: Expr[F1, A, Boolean]): Delete[F1, A] = Delete(table, expr)
 
-    def all[F]: Delete[Features.Literal, A] = Delete(table, Expr.literal(true))
+    def all[F1]: Delete[Features.Literal, A] = Delete(table, Expr.literal(true))
   }
 
   sealed case class Delete[F, A](table: Table.Aux[A], whereExpr: Expr[F, A, Boolean])
--- a/core/jvm/src/main/scala/zio/sql/delete.scala
+++ b/core/jvm/src/main/scala/zio/sql/delete.scala
@@ -2,12 +2,7 @@ package zio.sql
 
 trait DeleteModule { self: ExprModule with TableModule =>
 
-  sealed case class DeleteBuilder[F[_], A, B](table: Table.Aux[A]) {
-    def where[F1](expr: Expr[F1, A, Boolean]): Delete[F1, A] = Delete(table, expr)
-
-    def all[F1]: Delete[Features.Literal, A] = Delete(table, Expr.literal(true))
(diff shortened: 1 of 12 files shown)


postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -222,7 +158,92 @@ trait PostgresModule extends Jdbc { self =>
...
+  }
+
+  object PostgresFunctionDef {
+    val Sind = FunctionDef[Double, Double](FunctionName("sind"))
+  }
+
+  override def renderDelete(delete: self.Delete[_, _]): String = {
+    val renderer = Renderer()
+    renderer.buildDeleteString(delete)
+    renderer.render()
```
jczuchnowski, line 241: so if I did
```scala
val renderer = Renderer()
renderer.buildDeleteString(delete)
renderer.render()
renderer.buildReadString(read)
renderer.render()
```
I would get some invalid SQL in the end.
I think introducing this Renderer and two-step String rendering is overcomplicating things. Wouldn't top level `buildDeleteString` and `buildReadString` achieve essentially the same without the risk of invalid state? Or even implementing them directly in `renderDelete` and `renderRead`?
The common private functions can still be shared.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -6,14 +6,9 @@ import zio.sql.Jdbc
  */
 trait PostgresModule extends Jdbc { self =>
 
-  object PostgresFunctionDef {
-    val Sind = FunctionDef[Double, Double](FunctionName("sind"))
-  }
-
-  override def renderRead(read: self.Read[_]): String = {
-    val builder = new StringBuilder
+  sealed case class Renderer(builder: StringBuilder = new StringBuilder) {
```
jczuchnowski, line 9: Looks like `private final class` would be a better fit here than `sealed case class`.

core/jvm/src/main/scala/zio/sql/delete.scala
```diff
@@ -4,6 +4,8 @@ trait DeleteModule { self: ExprModule with TableModule =>
 
   sealed case class DeleteBuilder[F[_], A, B](table: Table.Aux[A]) {
     def where[F1](expr: Expr[F1, A, Boolean]): Delete[F1, A] = Delete(table, expr)
+
+    def all[F1]: Delete[Features.Literal, A] = Delete(table, Expr.literal(true))
```
jczuchnowski, line 8: So, this isn't really nice from the usability perspective. What if we removed `DeletBuilder` entirely, moved the `where` method to the `Delete` (just like `Select` has) and made `deleteFrom` return `Delete` right away (with `where` equal true)?

jczuchnowski:
We need to figure out some details but otherwise great work here.

core/jvm/src/main/scala/zio/sql/delete.scala
kidonm: Let's do that.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
kidonm: Good point. Will fix it.

jczuchnowski (approved):

Issue #247, opened 2020-11-18
Add test for 'Replace' function for PostgreSQL


jczuchnowski:

rpiotrow:
I would like to try solve this one.

angelicarao:
@rpiotrow Did you manage to solve this? I'm having an [issue](https://github.com/zio/zio-sql/issues/246#issuecomment-731577346) with my ticket and I feel it might affect yours too.

rpiotrow:
@angelicarao I made a MR, please take a look if it solves the problem you have encountered.

PR #285, opened 2020-11-21
Add test for 'Replace' function for PostgreSQL (and fix its type)

rpiotrow:
Addresses #247

 * **fix**: literals rendering 
 * **test**: more tests for select queries in PostgreSQL
 * **fix**: 'Replace' function type
 * **test**: add test for 'Replace' function for PostgreSQL
 * **fix**: update test schema

### DIFF
--- a/core/jvm/src/main/scala/zio/sql/expr.scala
+++ b/core/jvm/src/main/scala/zio/sql/expr.scala
@@ -4,8 +4,8 @@ import java.time._
 
 import scala.language.implicitConversions
 
-trait ExprModule extends NewtypesModule with TypeTagModule with FeaturesModule with OpsModule {
-  self: SelectModule with TableModule with TypeTagModule =>
+trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
+  self: SelectModule with TableModule =>
 
   /**
    * Models a function `A => B`.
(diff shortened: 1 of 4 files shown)


core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -108,7 +108,21 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
...
+        case TypeTag.TDouble    => Expr.Literal(a)
+        case TypeTag.TFloat     => Expr.Literal(a)
+        case TypeTag.TInt       => Expr.Literal(a)
+        case TypeTag.TLong      => Expr.Literal(a)
+        case TypeTag.TShort     => Expr.Literal(a)
+        case _                  => literalAsString(a)
+      }
+
+    private def literalAsString[A](a: A) =
+      Expr.Literal(s"'${a.toString}'").asInstanceOf[Expr[Features.Literal, Any, A]]
```
jczuchnowski, line 125: Generally it's a good direction, but I have the same doubts about it as in https://github.com/zio/zio-sql/pull/269.
This really fits better into rendering.

jczuchnowski: So TBH I would drop this change and make the test wrap the String params in single quotes for the moment. Ugly I know. But it's not a precedent. Someone will fix this in another PR in the renderer.

jczuchnowski:
Some coordination concern, but otherwise awesome work that was much needed.

core/jvm/src/main/scala/zio/sql/expr.scala
rpiotrow: I can change it in the afternoon (together with conflict resolving), but I will need to add some nasty workaround to tests added to `PostgresModuleTest.scala`.

rpiotrow:
Closed in favor of #319

PR #286, opened 2020-11-21
Add 'random' function to PostgresModule

### LINKED ISSUE #180: Add 'random' function to PostgresModule
Add support for function `random` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-math.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

riccardocorbella:
Fixes #180


riccardocorbella:
Random function need to be called using parentheses while current_date does not.
Screenshot 2020-11-23 at 08 48 33

jczuchnowski:
@riccardocorbella yes, exactly. And I see you've already introduced the special case for rendering. Although I think it's the `current_date` and `current_timestamp` should be the exception from the rule in PostgreSQL.

jczuchnowski (approved):

PR #290, opened 2020-11-21
Issue 214 and 205

### LINKED ISSUE #205: Add 'current_time' function to PostgresModule
Add support for function `current_time` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-datetime.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

### LINKED ISSUE #214: Add 'timeofday' function to PostgresModule
Add support for function `timeofday` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-datetime.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

antoniograndinetti:
Fixes #214 
Fixes #205


jczuchnowski:
@robmwalsh could you coordinate :pray: ? Solves similar problem to https://github.com/zio/zio-sql/pull/263

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -210,6 +214,9 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
 
   sealed case class FunctionDef[-A, +B](name: FunctionName) { self =>
 
+    def apply[Source, B1 >: B]()(implicit typeTag: TypeTag[B1]): Expr[Unit, Source, B1] =
```
robmwalsh, line 217: All the features that create a new `Expr` (rather than operating on an existing one) in some way introduce a new `Feature` type (that's the `F` parameter getting passed around everywhere). Can you please add a `Features.Function0` (like `Features.Source`) and use that instead of `Unit` for the `F` parameter?

As an `Expr[F, -A, +B]` models a function `A => B` (with some `Feature`s `F`) we can get rid of `Source` entirely and change the `A` (currently `Source`) to `Any` because a `Function0` can accept any value (it's ignoring everything in `Source`).

I think we should end up with something like 
def apply[B1 >: B]()(implicit typeTag: TypeTag[B1]): Expr[Features.Function0, Any, B1]

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -158,6 +158,10 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
       def typeTag: TypeTag[Z] = implicitly[TypeTag[Z]]
     }
 
+    sealed case class FunctionCall0[F, A, B, Z: TypeTag](function: FunctionDef[B, Z]) extends InvariantExpr[F, A, Z] {
```
robmwalsh, line 161: I think this should be 
```scala
sealed case class FunctionCall0[Z: TypeTag](function: FunctionDef[Any, Z]) extends InvariantExpr[Features.Function0, Any, Z]
```
since we don't care what the input type is (there is no input for a Function0) and we're introducing a new Feature that isn't based on anything else (see other comment re Features.Function0)

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -35,6 +37,37 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+      val query = select(Timeofday()) from customers
+
+      val testResult = execute(query).to[String, String](identity)
+
+      val assertion =
+        for {
+          r <- testResult.runCollect
+        } yield assert(r.head)(
+          matchesRegex(
+            "[A-Za-z]{3}\\s[A-Za-z]{3}\\s[0-9]{2}\\s(2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9].[0-9]{6}\\s[0-9]{4}\\s[A-Za-z]{3}"
```
robmwalsh, line 52: I'll have to take your word for this :)

robmwalsh (changes requested):
Awesome work here @antoniograndinetti! A few things to fix up, mainly relating to types. You've stumbled into some of the slightly more complicated type machinery here - I've tried to explain the relevant bits but feel free to ask if anything's unclear.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -1,58 +1,68 @@
...
+      case Expr.Literal(value)                                                  =>
         val _ = builder.append(value.toString) //todo fix escaping
-      case Expr.AggregationCall(param, aggregation)                     =>
+      case Expr.AggregationCall(param, aggregation)                             =>
         builder.append(aggregation.name.name)
         builder.append("(")
         buildExpr(param)
         val _ = builder.append(")")
-      case Expr.FunctionCall1(param, function)                          =>
+      case Expr.FunctionCall0(function) if function.name.name == "current_time" =>
```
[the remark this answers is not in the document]
antoniograndinetti: the problem here is that "timeofday" need empty curly brackets, while "current_time" doesn't... How can I distinguish these two cases?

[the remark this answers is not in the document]
robmwalsh: Ah, sorry I didn't notice that. That means we need another case to match against for things that require no parens. This shouldn't use FunctionDef as that means we have to call apply on it (which would mean adding parens to call it in our code), but should extend InvariantExpr directly like FunctionCall0 (which we still need for the ones with parens). Then your method definition would become something like
`val timeOfDay = ParenlessFunctionCall0[String]("time_of_day")`

core/jvm/src/main/scala/zio/sql/expr.scala
antoniograndinetti: Hello Rob, I'm having some issues applying this (all the others fixes have been applied). The next line (218) give me error. I attach a screenshot.
Screenshot 2020-11-27 at 18 18 57

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
antoniograndinetti: that's my idea to test timeofday, because this function return everytime a precise instant, so it will be impossible to assert with equality... right? So I try with regex matching these cases

core/jvm/src/main/scala/zio/sql/expr.scala
antoniograndinetti: At the moment I commit without this change and it works.

antoniograndinetti: Ok, delete all the previous comments, I think that I did it! Look at my last commit.. Thanks!

robmwalsh: you can cast here instead of using the type ascription for self (e.g. `self.asInstanceOf[Func`... instead of `self:Func`...)

antoniograndinetti: look at my last comment ;)

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -210,6 +226,10 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
 
   sealed case class FunctionDef[-A, +B](name: FunctionName) { self =>
 
+//    def apply[Source, B1 >: B]()(implicit typeTag: TypeTag[B1]): Expr[Unit, Source, B1] =
+    def apply[B1 >: B]()(implicit typeTag: TypeTag[B1]): Expr[Features.Function0, Any, B1] =
```
robmwalsh, line 230: can we add implicit `ev :  Any <:< A`? (see other apply methods). This will make sure only a FunctionCall0 can call apply with no arguments.

core/jvm/src/main/scala/zio/sql/expr.scala
robmwalsh: type changes? It shouldn't be possible to create a `FunctionCall0` that has features other than `Features.FunctionCall0` and an input type other than `Any`, and the `FunctionDef` needs to be our equivalent of `Any => Z`

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -35,6 +37,37 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+          )
+        )
+      assertion.mapErrorCause(cause => Cause.stackless(cause.untraced))
+    },
+    testM("current_time") {
+      val query = select(CurrentTime()) from customers
```
robmwalsh, line 58: :'(

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -158,6 +158,16 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
       def typeTag: TypeTag[Z] = implicitly[TypeTag[Z]]
     }
 
+    sealed case class ParenlessFunctionCall0[Z: TypeTag](function: ParenlessDef[Z])
```
robmwalsh, line 161: I think we just need a `FunctionName` for this, then you can use it directly without calling apply

robmwalsh (changes requested):
Nice work here @antoniograndinetti - just a few more things to go to make the parenless functions parenless in our DSL as well, and a bit of extra type safety

jczuchnowski (approved):

robmwalsh:
Awesome work here @antoniograndinetti! Thanks for your contribution

PR #291, opened 2020-11-21
166: fix compilation error in dotty

### LINKED ISSUE #166: Fix Dotty compilation errors
There are several compilation errors while using Dotty. Once they're fixed we'll be able to include Scala 3 check in build and Scala 3 release.
To compile with Dotty run:
```
sbt "++ 0.27.0-RC1" compile
```

scalavision:
Fixes #166 

This makes dotty compile and it also still compiles for scala 2.13.3.

The problem is that you need to add parenthesis in the ShoppingSchema code. I've refactored out the dotty example from the scala2 example.

To make the dsl like syntax work for scala 3 I think we need to add the @infix annotation, but not sure if this will work with scala 2, so maybe this could be added later if needed?


jczuchnowski:
Yes, fixing the syntax is another problem. Great job here! I can finally enable Scala 3 checks in CI :)

jczuchnowski:
Ah, wait. `test:compile` is still failing. Can you take a look?

scalavision:
I have looked a bit more into this. It seems that dotty is not able to handle the deeply nested typelevel structure built using the `++` operator (if I understand things correctly here):

* https://github.com/zio/zio-sql/blob/master/core/jvm/src/main/scala/zio/sql/table.scala#L16

I have basically circumvented that by explicitly calling everything using parenthesis, thus I guess that dotty is able to infer this correctly when helped out. Unfortunately, this breaks the nice dsl like syntax of the sql.

I suspect this being a limitation in the dotty compiler, so maybe file an issue for this in dotty? I think dotty should be able to handle this. Alternatively, we could try to exclude the test sources for dotty, but I am a bit short on sbt magic, and it is not a solution in the long run anyways.

The assumption done earlier on infix annotation is wrong I think, because symbolic operators are supposed to be handled exactly as in scala 2.

jczuchnowski:
Is it possible it's dotty version we're using? We're stuck on `0.27.0-RC1` because of ZIO. Maybe on `3.0.0-M2` it'll get better. We could recheck with your PR https://github.com/zio/zio/pull/4483

scalavision:
Thanks @jczuchnowski , compiling against scala 3.0.0-M2 and the branch (just published things locally) produces the same problems it seems. I'll make another go, and try to isolate the problems after release of new version of zio. I have made a PR with the changes that I think is compatible also with scala 2 here:

https://github.com/zio/zio-sql/pull/343

Suggest closing this and see if we can find a better solution later :-)

jczuchnowski:
@scalavision ok, let's do that. I'll do some experiments with 3.0.0-M2 too.

Issue #211, opened 2020-11-15
Add 'make_*' functions to PostgresModule


jczuchnowski:
Add support for functions `make_date`, `make_interval`, `make_time`, `make_timestamp`, `make_timestampz` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-datetime.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

brbrown25:
I can pick this up

PR #295, opened 2020-11-21
#211 Add 'make_*' functions to PostgresModule

brbrown25:


postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -35,6 +38,59 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+      val expected   = LocalTime.parse("08:15:23.5")
+      val testResult = execute(query).to[LocalTime, LocalTime](identity)
+
+      val assertion = for {
+        r <- testResult.runCollect
+      } yield assert(r.head)(equalTo(expected))
+
+      assertion.mapErrorCause(cause => Cause.stackless(cause.untraced))
+    },
+    testM("make_timestamp") {
```
brbrown25, line 71: Both this and the make_timestampz complain that there are two many arguments, I'm not really sure what's going on there. Also not really sure how to tackle the case of interval which takes a format of "<interval> => <value>) Maybe a custom parser for that?

robmwalsh: RE the timestamps, in ExprModule.Expr there are FunctionCall1 through 4 and corresponding overloaded apply methods on FunctionDef. You've got 6 arguments here, so you need a FunctionCall6 and an apply method on FunctionDef that can make one. Bonus points for 5 while you're at it ;). As a side note, I don't think the options are necessary here as I think they're all required. If not you could just use default values (0) for all

for make interval, you could use default values for all the arguments which would give a nice API. I think you're up for FunctionCall7 as well...

brbrown25: @robmwalsh thanks so much for the help with this! I have to step away for the evening but will try and get this wrapped up by tomorrow!

brbrown25:
- [ ] sync with master
- [ ] implement interval
- [ ] add interval test
- [ ] ensure tests are working

brbrown25:
I'll be hacking away at this again today, got side tracked with the holidays!

brbrown25:
@robmwalsh I've finally got this working, I'm thinking it's actually using the toString version of the Interval and Timestampz case classes which might be fine for now. Let me know any thoughts you have for improving and I'll address them. I added in the postgres driver to reference the PGInterval class, which is similar to how slick implements this functionality.

robmwalsh (changes requested):
nice work - I don't have a problem with adding the postgres driver as I think we're going to move away from JDBC. A few more changes and this should be good to go. Let me know if you need a hand with any of it.

core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -193,6 +192,160 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
...
+              ]
+            ]
+          ],
+          A,
+          Z
+        ] {
+      def typeTag: TypeTag[Z] = implicitly[TypeTag[Z]]
+    }
+
+    sealed case class FunctionCall10[
```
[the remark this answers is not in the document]
brbrown25: Yeah that was just me going with gusto and adding it, I can remove it! Totally in agreement about code gen for large bodies.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -34,6 +34,8 @@ trait TypeTagModule {
     implicit case object TString                                                  extends NotNull[String]
     implicit case object TUUID                                                    extends NotNull[UUID]
     implicit case object TZonedDateTime                                           extends NotNull[ZonedDateTime]
+    implicit case object TInterval                                                extends NotNull[Interval]
+    implicit case object TTimestampz                                              extends NotNull[Timestampz]
```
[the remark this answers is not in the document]
brbrown25: Yeah that makes sense I was wondering about the best way to extend that. Will make that change!

core/jvm/src/main/scala/zio/sql/custom/Interval.scala
```diff
@@ -0,0 +1,94 @@
+package zio.sql.custom
```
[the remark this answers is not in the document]
brbrown25: Yeah I'll swizzle that into there and rearrange things accordingly.

brbrown25:
- [ ] Move Interval.scala from `zio.sql.custom` -> `zio.sql.postgresql`
- [ ] Create `PostgresTypeTagExtension` and move PG specifics into that
- [ ] Remove any Arrity FunctionCallX that isn't really used
- [ ] Double check which modules I'm adding the pg driver dependency to

brbrown25:
@robmwalsh this is a little bit closer to what I think we'd like ultimately. I'm wondering What you think I need to either 1) just us the toString I've defined on Timestampz and Interval or 2) have it use the tupled form of them. Also I think you were suggesting I remove the Timestampz case class, given how I've structured this do you think that still makes sense. My main motivator was that it provides a nice interface for creating them that matches the function args of the postgres function. Excited to get this a little bit closer :)

brbrown25:
@robmwalsh I have the question above about Timestampz and wondering if you have any ideas about how to fix my broken tests.

robmwalsh:
@brbrown25 @jczuchnowski we now have a working dialect specific type tag for postgres :)

The main issue was we were conflating type tags (which are associated with the values but are just implicit objects) with the values themselves. `Interval` didn't need to extend anything.

Thanks to @jczuchnowski for finding the cause of the compiling never ending.

@brbrown25 can you please pull the latest commits? your tests are now failing again :)
Interestingly there's an error involving the postgres library objects - I don't think we should be using `PGInterval` directly, just providing a way to convert to and from one to our `Interval` representation. I haven't actually looked at any of your test code though so could be something simple.

brbrown25:
> @brbrown25 @jczuchnowski we now have a working dialect specific type tag for postgres :)
> 
> 
> 
> The main issue was we were conflating type tags (which are associated with the values but are just implicit objects) with the values themselves. `Interval` didn't need to extend anything.
> 
> 
> 
> Thanks to @jczuchnowski for finding the cause of the compiling never ending.
> 
> 
> 
> @brbrown25 can you please pull the latest commits? your tests are now failing again :)
> 
> Interestingly there's an error involving the postgres library objects - I don't think we should be using `PGInterval` directly, just providing a way to convert to and from one to our `Interval` representation. I haven't actually looked at any of your test code though so could be something simple.

Thanks! Yeah I'll pull and look. The PGInterval should only be used when the decoding is happening but I'll get to the bottom of that!

brbrown25:
[] pull latest
[] fix failing test

brbrown25:
@robmwalsh @jczuchnowski I'm down to the following issues that I'm not quite sure how to address.
1) The make_timestamptz I want to render as `SELECT make_timestamptz(2013, 7, 15, 8, 15, 23, '+00')` but currently it's rendering as `SELECT make_timestamptz('2013-07-15T08:15:23.5+00:00')` which is it just stringing the ZonedDatetime. So I switched to just passing in a tuple of the args and that makes it pass for now. One issue with the ZonedDateTime is that you can't pass in seconds as a decimal so 23.5 is a valid value to the make_timestamptz for seconds, however ZonedDateTime wants everything as ints.
2) The interval test is now throwing
```
[info]     An unchecked error was produced.
[info]     scala.NotImplementedError: an implementation is missing
```
But I'm not quite sure where that's coming from.

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -1066,6 +1067,92 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+
+      (for {
+        t1 <- assertM(runTest(Interval()))(equalTo(Interval()))
+        t2 <- assertM(runTest(Interval(days = 10)))(equalTo(Interval(days = 10)))
+        t3 <- assertM(
+                runTest(Interval(years = 10, months = 2, weeks = 1, days = 5, hours = 6, minutes = 20, seconds = 15))
+              )(
+                equalTo(Interval(years = 10, months = 2, weeks = 1, days = 5, hours = 6, minutes = 20, seconds = 15))
+              )
+      } yield t1 && t2 && t3).mapErrorCause(cause => Cause.stackless(cause.untraced))
```
robmwalsh, line 1111: this `.mapErrorCause(cause => Cause.stackless(cause.untraced))` is throwing away the tracing information, hence the `[info]     No ZIO Trace available.`. comment this out and it'll show the full stack trace in the test results (which is unideal, and end up doing this - this will be fixed in 2.x I believe)

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -34,7 +34,9 @@ trait TypeTagModule {
     implicit case object TString                                                  extends NotNull[String]
     implicit case object TUUID                                                    extends NotNull[UUID]
     implicit case object TZonedDateTime                                           extends NotNull[ZonedDateTime]
-    sealed case class TDialectSpecific[+A](typeTagExtension: TypeTagExtension[A]) extends NotNull[A]
+    sealed case class TDialectSpecific[+A](typeTagExtension: TypeTagExtension[A]) extends NotNull[A] {
+      def decode[A]: Either[Int, String] => A = ???
```
robmwalsh, line 38: I'd hazard a guess that this is your problem. I'm not sure exactly what you're trying to do here but keep in mind this has to work with all custom data types from any dialect.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -128,69 +267,72 @@ trait PostgresModule extends Jdbc { self =>
...
+        case TBoolean             => render(lit.value)           //default toString is probably ok
+        case TDouble              => render(lit.value)           //default toString is probably ok
+        case TFloat               => render(lit.value)           //default toString is probably ok
+        case TInt                 => render(lit.value)           //default toString is probably ok
+        case TLong                => render(lit.value)           //default toString is probably ok
+        case TShort               => render(lit.value)           //default toString is probably ok
+        case TString              => render("'", lit.value, "'") //todo fix escaping
+        case TDialectSpecific(tt) =>
+          tt match {
+            case tt @ TInterval => render(tt.cast(lit.value))
```
robmwalsh, line 281: btw this is where you can render however you like. `TString` is probably a decent example here.

core/jvm/src/main/scala/zio/sql/typetag.scala
brbrown25: Ah yes I see what the issue is, will have a fix on this shortl!

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -12,9 +13,21 @@ trait PostgresModule extends Jdbc { self =>
...
-      implicit case object TInterval extends PostgresTypeTag[Interval]
+      implicit case object TInterval extends PostgresTypeTag[Interval] {
+        override def decode[Interval, DecodingError](
+          column: Either[Int, String],
+          resultSet: ResultSet
+        ): Either[DecodingError, Interval] =
+          scala.util
+            .Try(Interval.fromPgInterval(new PGInterval(column.fold(resultSet.getString(_), resultSet.getString(_)))))
+            .fold(
+              _ => Left(DecodingError.UnexpectedNull(column).asInstanceOf[DecodingError]),
```
brbrown25, line 27: I had to add the `.asInstanceOf` to make the compiler happy with this

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -29,10 +42,11 @@ trait PostgresModule extends Jdbc { self =>
       (year, month, day, hour, minute, second, timeZone)
 
     //Based upon https://github.com/tminglei/slick-pg/blob/master/src/main/scala/com/github/tminglei/slickpg/PgDateSupport.scala
+    //The postgres make_interval function allows for specifying weeks, but that's not supported in the PGInterval class, so we
```
brbrown25, line 45: upon investigation it appears that weeks is not supported in the PGInterval class, so even I've dropped it since we wouldn't be able to handle it at the moment. It might make sense to actually remove the PG compile time dependency and just implement a string parser to convert into our internal interval class. If you like that idea I can make a separate issue and implement it.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -77,4 +79,6 @@ trait TypeTagModule {
     implicit case object TDoubleIsNumeric        extends AbstractIsNumeric[Double]
     implicit case object TBigDecimalIsNumeric    extends AbstractIsNumeric[BigDecimal]
   }
+
+//  type PostgresTypeTagExtension
```
robmwalsh, line 83: this can be deleted

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -1,17 +1,20 @@
...
 
-  type TypeTagExtension[+A]
+  type TypeTagExtension[+A] <: Tag[A]
 
-  sealed trait TypeTag[+A] {
-    private[zio] def cast(a: Any): A = a.asInstanceOf[A]
+  trait Tag[+A] {
+    private[zio] def cast(a: Any): A                                                  = a.asInstanceOf[A]
+    //This is required when implementing database specific custom types.
+    def decode[A, E](column: Either[Int, String], resultSet: ResultSet): Either[E, A] = ???
```
robmwalsh, line 15: this doesn't fit here because all the standard type tags don't need this. If someone calls this on one of the standard types, it goes boom. See if you can find somewhere else for it that won't risk a runtime error.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -1,17 +1,20 @@
...
+import java.sql.ResultSet
 import java.time._
 import java.util.UUID
-
 import zio.Chunk
 
 trait TypeTagModule {
 
-  type TypeTagExtension[+A]
+  type TypeTagExtension[+A] <: Tag[A]
```
robmwalsh, line 10: hint: Tag[A] with Decodeable[E, A]? not sure if we need the E, won't this always be a DecodeError?

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
robmwalsh: I do like this idea. Please create that issue and delete this comment (not the attribution though!).

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
robmwalsh: Odd. I don't think this should be required, but the cast looks sound.

robmwalsh (changes requested):
very minor stuff now :) Please also do a final pass to make sure you remove any unneeded comments

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
brbrown25: Sweet! I've created https://github.com/zio/zio-sql/issues/369 and can pick that up next.

core/jvm/src/main/scala/zio/sql/typetag.scala
brbrown25: yeah that's a good shout out.

brbrown25:
@robmwalsh I've gone ahead and cleaned this up and squashed it down into a single commit. I really wanted to be able to do something like `def decode[A, E <: DecodingError](column: Either[Int, String], resultSet: ResultSet): Either[E, A] = ???` for the definition of decode, but the compiler wouldn't let me. I also made one more change which was bringing back the Timestampz case class. Let me know any additional changes you'd like and I'll rock them out quickly. I should have the PGInterval pr in the next week or so.

robmwalsh (changes requested):
Thanks, just some (hopefully) minor stuff now.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -1,17 +1,22 @@
...
 import java.util.UUID
-
 import zio.Chunk
 
 trait TypeTagModule {
 
-  type TypeTagExtension[+A]
+  type TypeTagExtension[+A] <: Tag[A] with Decodeable[A]
+
+  trait Decodeable[+A] {
```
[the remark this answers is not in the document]
brbrown25: [] fix typo

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -1,17 +1,22 @@
...
+    def decode[A, DecodingError](column: Either[Int, String], resultSet: ResultSet): Either[DecodingError, A] = ???
```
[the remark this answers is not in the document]
brbrown25: [] drop the ???

robmwalsh (approved):
boom!

PR #301, opened 2020-11-21
transactions

### LINKED ISSUE #221: Add support for transactions
It should be possible to make several SQL statements in one transaction (e.g several inserts to different tables).

It would nice if transactions are orthogonal to SQL statements. Assuming layered application with services and repositories, it would be nice to control scope (and maybe isolation level of the transaction) in the service (which can invoke many repositories which does not know anything about transactions).

fokot:
Fixes #221

### DIFF
--- a/jdbc/src/main/scala/zio/sql/transaction.scala
+++ b/jdbc/src/main/scala/zio/sql/transaction.scala
@@ -1,8 +1,8 @@
 package zio.sql
 
-import zio.{Cause, ZIO}
+import zio.{ Cause, ZIO }
 
-trait TransactionModule { self : SelectModule with DeleteModule with UpdateModule =>
+trait TransactionModule { self: SelectModule with DeleteModule with UpdateModule =>
 
   import Transaction._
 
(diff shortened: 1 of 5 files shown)


jczuchnowski:
@fokot I took a first look at it. Looks pretty awesome! I'll make one more pass tomorrow with a fresh eyes though ;).

jdbc/src/main/scala/zio/sql/transaction.scala
```diff
@@ -0,0 +1,66 @@
+package zio.sql
+
+import zio.{ Cause, ZIO }
+
+trait TransactionModule { self: SelectModule with DeleteModule with UpdateModule =>
+
+  import Transaction._
+
+  sealed trait Transaction[-R, +E, +A] { self =>
```
jczuchnowski, line 9: Do you think it makes sense to introduce the same pattern as with all ZIO-like effects - `ZTransation[R, E, A]` and an alias `Transaction[E, A]`?

jczuchnowski:
@fokot just one small comment, but this looks very good and it's a great progress. 
As we've talked, once we have delete and insert in place, we'll be able to do more comprehensive tests and examples, but for now it's good.
If you could just resolve the conflicts :pray: .

jczuchnowski (approved):

jdbc/src/main/scala/zio/sql/transaction.scala
fokot: I will resolve it in next PR

Issue #208, opened 2020-11-15
Add 'extract' function to PostgresModule


jczuchnowski:

rpiotrow:
I'll try this one.

jczuchnowski:
This is still an open issue.

jczuchnowski:
This function needs extra handling as it's input is not a simple list of parameters. The signature is
```
extract ( field from timestamp ) 
```

PR #302, opened 2020-11-21
WIP: Add 'extract' function to PostgresModule

rpiotrow:
### Changes

 * add optional delimiter to FunctionDef for function call with 2 params
   since extract has form "extract(param1 from param2)"
 * add `TimestampPart` and corresponding type tags that are dedicated
   to PostgreSQL

### Still to do

 * make extract function available for other date like types
   (e.g. LocalDateTime, Instant, etc.), now it works only for LocalDate

Addresses #208

### DIFF
--- a/core/jvm/src/main/scala/zio/sql/expr.scala
+++ b/core/jvm/src/main/scala/zio/sql/expr.scala
@@ -166,9 +166,18 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
     sealed case class FunctionCall2[F1, F2, A, B, C, Z: TypeTag](
       param1: Expr[F1, A, B],
       param2: Expr[F2, A, C],
+      paramDelimiter: String,
       function: FunctionDef[(B, C), Z]
     ) extends InvariantExpr[Features.Union[F1, F2], A, Z] {
       def typeTag: TypeTag[Z] = implicitly[TypeTag[Z]]
+
+    }
+    object FunctionCall2 {
+      def apply[F1, F2, A, B, C, Z: TypeTag](
+        param1: Expr[F1, A, B],
+        param2: Expr[F2, A, C],
+        function: FunctionDef[(B, C), Z]
+      ): FunctionCall2[F1, F2, A, B, C, Z] = new FunctionCall2(param1, param2, ",", function)
     }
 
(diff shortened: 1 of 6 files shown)


core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -246,45 +259,60 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
     }
   }
 
-  object FunctionDef {
+  sealed case class FunctionDefWithDelimiter[-A, +B](name: FunctionName, delimiter: String) extends FunctionDef[A, B] {
```
jdegoes, line 262: What is the purpose of `delimiter`? Can you help me understand this a bit more?

rpiotrow: Standard function in SQL looks like this: `SELECT func(param1, param2)` e.g. `SELECT date_part('minutes', source)` while `extract` is different, it takes two parameters, but it looks like this: `SELECT extract(minutes from source)`. So for "standard" function delimiter is `,` while for extract function it is ` FROM `.

rpiotrow:
I think alternative solution to functions with custom argument passing is here: https://github.com/zio/zio-sql/pull/309

core/jvm/src/main/scala/zio/sql/expr.scala
jczuchnowski: This looks like a very special case. Maybe we could solve it at the `renderRead` level?

rpiotrow: You mean something like idea in #309 (extension to `Expr` with custom renderer for it) or just simple `if` checking the name of the function in the `renderRead` ;) ?

rpiotrow:
Any idea how to solve problem in "TODO":
> make extract function available for other date like types (e.g. LocalDateTime, Instant, etc.)

robmwalsh:
@rpiotrow using #309 you could add a custom expr that requires a certain context bound. #271 is a general solution to the problem for simple functions, but since you need custom rendering as well I don't think this would work in this case.

jczuchnowski:
No activity for some time. I'm closing this for now and we'll revisit the issue after the first release.

Issue #191, opened 2020-11-14
Add 'pg_client_encoding' function to PostgresModule


jczuchnowski:
Add support for function `pg_client_encoding` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-string.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

bonczek:
I'll take that

PR #304, opened 2020-11-21
Add 'pg_client_encoding' function to PostgresModule #191

bonczek:
That's my first PR ever on open source project, so just let me know if I messed up something obvious. My concerns:
1. FunctionCall0 was already implemented, so I rebased it on changes here: https://github.com/zio/zio-sql/pull/286/
2. I followed similar design like in Random(), but honestly I'm not convinced to using `select(PgClientEncoding()) from customers`. IMO it should be fine to just use it without 'from': `SELECT pg_client_encoding();` but I had no idea how to achieve that.


jczuchnowski:
Looks good :+1: 
As for your concern - you're right. You shouldn't be forced to do `from customers`. But for now the DSL doesn't allow for it. I'll create a ticket for that.

jczuchnowski:
I'll merge it after `random` gets in.

jczuchnowski (approved):

Issue #308, opened 2020-11-21
support custom renderings for functions with odd syntax


robmwalsh:
discovered by @marcinkrykowski in #245 

e.g. overlay needs specific keywords inserted between arguments

`overlay(<main_string> placing <replacing_string> from <starting_position> [ for <mumber_of_characters>] )`

I'll be working on this one

jczuchnowski:
This should solve #601 and help #208

PR #309, opened 2020-11-21
support dialect specific `Expr`s

robmwalsh:
#308


postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -11,6 +12,11 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
   import this.FunctionDef._
 
   val spec = suite("Postgres FunctionDef")(
+    test("custom expr rendering") {
+      val query  = select(Overlay("w333333rce", "resou", 3, 5)) from customers
+      val render = renderRead(query) //SELECT overlay(w333333rce placing resou from 3 for 5) FROM customers
+      assert(render)(equalTo("SELECT overlay('w333333rce' placing 'resou' from 3 for 5) FROM customers"))
+    } @@ ignore, //todo fix rendering of strings
```
pjazdzewski1990, line 19: Wouldn't it be enough to just do
`implicit def literal(str: String): Expr[Features.Literal, Any, String] = Expr.Literal(s"'${str}'")`
in expr?

robmwalsh: this would break once string rendering is fixed - please see https://github.com/zio/zio-sql/pull/307#discussion_r528286267

robmwalsh:
I had some discussion with @fokot on this last night - we thought that it might be good to ensure that any provided dialect specific expr should be capable of rendering any introduced types itself. The renderer should delegate to the custom expr renderer whenever it encounters such a type, and the custom expr should return back to the main renderer for any standard types. This change would leave the door open to potentially allowing users to add custom functions such as those provided by postgres plugins... it would also potentially allow users to open the door to sql injection etc. if they didn't use it correctly, but I think the benefits outweigh the risks as long as we slap sufficiently large warning labels on the box. @jdegoes @jczuchnowski what are your thoughts on this?

fokot:
This is how it was in quill https://getquill.io/#extending-quill-infix and I liked it a lot. Infix parameters are escaped, so it didn't even lead to sql injection on them. Slick is way more limited in customisation https://scala-slick.org/doc/3.2.0/userdefined.html#scalar-database-functions. I could map and use my stored functions but not much more.

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -2,12 +2,36 @@ package zio.sql.postgresql
...
+      replacing: Expr[F2, Source, String],
+      starting: Expr[F3, Source, Int],
+      number: Expr[F4, Source, Int]
+    ) extends PostgresExprExtension[F1 :||: F2 :||: F3 :||: F4, Source, String] {
+      val functionName = "overlay"
+    }
+  }
+  object PostgresFunctionDef   {
+    val Sind    = FunctionDef[Double, Double](FunctionName("sind"))
+    val Overlay = PostgresExprExtension.Overlay
```
jczuchnowski, line 34: Isn't this a problem that there's no indication that Overlay is a FunctionDef?

robmwalsh: It's not a `FunctionDef`, at least in terms of how `FunctionDef` is defined. `Expr` is our concept of a function A => B with some features F, `FunctionDef` is just a syntax class to allow you easily create `FunctionN` case classes for simple functions, which are themselves just extending `InvariantExpr`

robmwalsh: a couple of options here:
1. simply rename `PostgresExprExtension` to `PostgresFunction` (since Expr and Function are both functions anyway)
2. we could maybe make `FunctionCall`N traits (not sealed) which could  be self typed as `Expr`s (making it effectively still sealed). We'd then have to come change the existing `FunctionCall`N classes to something like `BasicFunctionCall2 extends InvariantExpr[F1 :||: F2], A, Z] with FunctionCall2[F1, F2 , A, Z] and then we could upcast all the return types to `FunctionCall`N for a nice standardised list of types in autocomplete - the user wouldn't even know which ones are dialect specific... I'm not sure if this is a good thing or not

jczuchnowski:
No activity for some time. I'm closing this for now and we'll revisit the issue after the first release.

Issue #311, opened 2020-11-22
Fix rendering of literals - Postgres module


robmwalsh:
Rendering of literals is broken for all modules - it's currently just using `.toString` on the underlying values. Using the type tag in the Expr.Literal, we need to determine the type of the enclosed `value` and render it using the provided builder. Bonus points for escaping to prevent SQL injection, but there's another ticket (#219) for that so just something that works and assumes benevolent users will do for now.

I've just opened one ticket for postgres - most of this work should be able to be copied between modules and adjusted rather than reinventing the wheel. I'll create tickets for the other modules once this one is complete.

```scala
override def renderRead(read: self.Read[_]): String = {
    ...
    def buildExpr[A, B](expr: self.Expr[_, A, B]): Unit = expr match {
      case Expr.Source(tableName, column)                               =>
     ...
      case Expr.Literal(value)                                          =>
           val _ = builder.append(value.toString) //<------this doesn't work real well D:
    }
  ...
}

robmwalsh:
I'll take this

PR #317, opened 2020-11-22
Render literals & break up giant `renderRead` function to allow reuse of e.g. `renderExpr` across renderings

robmwalsh:
not a complete solution but at least fixes string rendering and should make rendering other types easier (at least shows the pattern)

partial solution to #311 

also made the rendering functions a bit more user friendly and modular


jczuchnowski:
@robmwalsh this is so much needed :pray: . Could you resolve the conflicts? Unfortunately there were couple of PRs that also modified the rendering.

robmwalsh:
really wish I did this one before the hackathon! It was interesting to see all the hacks people added to get their tests working... I broke a couple of tests which are now `@@ ignore`d but I think they need custom rendering, so can fix once #309 is merged.

btw I've noticed that `fmt` frequently doesn't do everything it needs to in one pass, so now it runs twice :) the 2nd run is < 1 second if it has nothing to do. I think that's a small price to pay for not having to do another commit/push/spin up CI.

jczuchnowski:
No, it's not fmt, it's me :). Sometimes when I resolve conflicts when merging master to someone else's branch, I make mistakes and then the check fails. But I merge anyway, because it would take forever to contact eveyone.
It's just temporary for the hackathon tickets.
Unless it's something else you have in mind.

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -854,8 +872,6 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
       } yield assert(r.head)(equalTo("UTF8"))
 
       assertion.mapErrorCause(cause => Cause.stackless(cause.untraced))
-    }
-  )
-
-  private def postgresStringEscape(s: String): String = s""" '${s}' """
+    } @@ ignore  //todo fix - select(PgClientEncoding())?
```
jczuchnowski, line 875: `from smth` is required now. We need to rethink `Read`.

robmwalsh:
This is a bit different, I merged master, ran `fmt`, refeshed IJ to make sure it picked up the changes, committed, and my branch failed the check. 

This stems from the fact that scalafmt isn't entirely idempotent. There are a few issues floating about but they keep closing them because it's a problem with one of their dependencies (e.g. https://github.com/scalameta/scalafmt/issues/917).  

Can we please merge this now? The ones that I broke will be resolved with the changes I requested here: https://github.com/zio/zio-sql/pull/290#discussion_r531206389

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
robmwalsh: I raised this #323 for this one. I think we should rename `Read.Select` to `Read.SelectFrom` and `SelectBuilder` to `Read.Select`, both extending a common trait which the renderer can match on.

I think the underlying issue with this one I think is just that it needs parens and other `Function0`s don't.

This is one of the changes I was talking about in my comment on https://github.com/zio/zio-sql/pull/290#discussion_r531206389

jczuchnowski (approved):

PR #321, opened 2020-11-24
Full stack mysql test environment

### LINKED ISSUE #69: Add full-stack test for MySQL
Depends on https://github.com/zio/zio-sql/issues/149

We need a full-stack test for MySQL. 

We need an automated ZIO Test that, when run, will execute a simple query, such as `select 1`, against a live-running MySQL, and verify the result set returned from the query.

This full-stack test will be the foundation for adding other more complicated tests.

This can follow the established pattern from `PostgresModuleTest`.

andrewhj:
Fixes #69 

Created mysql environment heavily insipred (copy/pasted) by the postgres
module.

MysqlModule.scala: performs the conversion of zio-sql functions into
mysql compatible sql.  Currently a copy/paste & rename of the postgres
module.

FunctionDefSpec.scala: module for testing mysql functions (lower, sin,
etc.)

MysqlModuleTest.scala: runs mysql tests, table selects, limits, offset,
etc.

MysqlRunnableSpec.scala:

ShopSchema.scala: defines schema (loaded in test container definition)


mysql/src/test/resources/shop_schema.sql
```diff
@@ -0,0 +1,84 @@
+create table simple
```
robmwalsh, line 1: this can be removed

robmwalsh, line 1: this can be removed

mysql/src/test/resources/shop_schema.sql
```diff
@@ -0,0 +1,84 @@
...
+
+create table products
+(
+    id varchar(36) not null primary key,
+    name varchar(255),
+    description varchar(255) not null,
+    image_url varchar(255)
+);
+
+insert into simple
```
robmwalsh, line 31: this can be removed

mysql/src/test/scala/zio/sql/mysql/MysqlModuleTest.scala
```diff
@@ -0,0 +1,54 @@
...
+//
+//      val testResult = execute(query).to[Integer, Integer](identity)
+//
+//      val assertion = for {
+//        r <- testResult.runCollect
+//      } yield assert(r)(hasSameElementsDistinct(expected))
+//
+//      assertion.mapErrorCause(cause => Cause.stackless(cause.untraced))
+//    },
+    testM("Can select from single table (simple)") {
```
robmwalsh, line 30: can we use an existing table like customers here?

mysql/src/test/scala/zio/sql/mysql/ShopSchema.scala
```diff
@@ -0,0 +1,48 @@
+package zio.sql.mysql
+
+import zio.sql.Jdbc
+
+trait ShopSchema extends Jdbc { self =>
+  import self.ColumnSet._
+
+  object Simple {
```
robmwalsh, line 8: this can be removed

robmwalsh (changes requested):
Awesome work thanks @andrewhj! Noting my comments here https://github.com/zio/zio-sql/issues/69#issuecomment-733430991 can you please remove the "simple" table and use one of the existing tables instead? Other than that I think this is good to merge!

andrewhj:
yes I can do that.  I added the simple table because the task started expanding further away from the original intent.  Mysql either doesn't support (or supports differently) the concept of uuid for example so I changed the id columns to varchar.  Is that ok?  Going in with that in mind, I'll work on matching up the rows and tests to postgres data example.

robmwalsh:
Varchar is horribly inefficient for storing UUIDs. Perhaps changing them to Ints for now would be better? A bit of a pain on the test data but find replace on the primary column wouldn't be amazingly terrible. I think down the track we should probably have something more like binary(16) as described here https://mysqlserverteam.com/storing-uuid-values-in-mysql-tables/ and take care of the conversion. That would be transparent when using zio-sql for queries but makes hand writing queries much harder. Might be a good justification for allowing users to define how they want to store types.

andrewhj:
well, that was a mistake.  I'll work through the merge issues.  I had tests working with varchar storing UUID.  I think that is a nice spot despite the mentioned inefficiency of varchar.  The tests remain closer to postgresql examples.  If we really want to convert to using ints instead of varchar/uuid, I can work on that next.   It requires changes to ShopSchema, shop_schema.sql, and the MysqlRunnableSpec.

jczuchnowski (approved):

Issue #216, opened 2020-11-15
deleteFrom is failing for Scala 2.12


jczuchnowski:
Trying an example delete in Examples.scala fails to compile for Scala 2.12:
```scala
val basicDelete = deleteFrom(users).where(fName === "Terrence")
```

kidonm:
I would like to look at this one.

PR #331, opened 2020-11-27
Make deleteFrom compile on 2.12

kidonm:
fix: https://github.com/zio/zio-sql/issues/216

Moving `Repr[_]` out of type member to a type parameter fixes `2.12` compilation.


core/jvm/src/main/scala/zio/sql/table.scala
```diff
@@ -135,8 +134,8 @@ trait TableModule { self: ExprModule with SelectModule =>
     trait Insanity {
       def ahhhhhhhhhhhhh[A]: A
     }
-    sealed trait Source extends Table with Insanity {
-      type Repr[_]
+    // NOTE: Moving Repr[_] from a type parameter to a type member breaks deleteFrom compilation on 2.12
+    sealed trait Source[Repr[_]] extends Table with Insanity {
```
robmwalsh, line 138: can you please remove the insanity trait and see if this change has fixed the "//The outer reference in this type test cannot be checked at run time?!" comments in the rendering?

robmwalsh (commented):
nice work @kidonm, I'm going to defer to @jdegoes on this one. We're deliberately forgetting information to prevent people constructing columns that don't exist in tables, and I think making this a type member may have been done intentionally. It used to be something like this (Repr was on Table rather than Table.Source) and it was changed to the current encoding.

core/jvm/src/main/scala/zio/sql/table.scala
kidonm: @robmwalsh the error is still present (not sure if relevant, changing `Source` to `sealed abstract class` makes the warning go away).

robmwalsh: Great, thanks for looking into that I don't think there's a reason Source needs to be a trait vs sealed abstract class, so might look to do that and remove `Insanity`. Will await John's verdict on this PR and take it from there.

jczuchnowski:
I had a session with @jdegoes about this problem. He came with a simple and elegant solution to this. Here's the PR https://github.com/zio/zio-sql/pull/403

jczuchnowski:
I'm closing this PR. @kidonm @robmwalsh big thanks for the work you did here. Sorry it took so long to resolve.

Issue #248, opened 2020-11-18
Add test for 'Rtrim' function for PostgreSQL


jczuchnowski:

visar:
I would like to tackle this.

Issue #243, opened 2020-11-18
Add test for 'Ltrim' function for PostgreSQL


jczuchnowski:

visar:
I would like to tackle this.

PR #339, opened 2020-12-08
Add test for ltrim & rtim

marekklis:
Implementation for: https://github.com/zio/zio-sql/issues/243 and https://github.com/zio/zio-sql/issues/248

BTW Postgres `ltrim` and `rtrim` have optional second parameter for defining what characters must be removed (space is a default value). More details: https://www.postgresql.org/docs/13/functions-string.html


jczuchnowski:
Good point.  I'll make a ticket for that.

PR #343, opened 2020-12-09
add missing dependencies to module pattern

scalavision:
I am not entirely sure why, but Scala 3 needs these modules to be available in order to compile. Strange that Scala 2 does not also require them :-) It could of course also be a bug in Scala 3 so take this with a grain of salt :-)


jczuchnowski:
Yes, I've seen this when switched to dotty. Weird.

jczuchnowski (approved):

PR #348, opened 2020-12-14
Add 'encode' and 'decode' functions to Postgres module

danieletorelli:
This is adding support for:
- [Encode](https://www.postgresql.org/docs/current/functions-binarystring.html#FUNCTION-ENCODE) function
- [Decode](https://www.postgresql.org/docs/current/functions-binarystring.html#FUNCTION-DECODE) function


postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -49,6 +49,8 @@ trait PostgresModule extends Jdbc { self =>
     val Now                         = FunctionDef[Any, ZonedDateTime](FunctionName("now"))
     val StatementTimestamp          = FunctionDef[Any, ZonedDateTime](FunctionName("statement_timestamp"))
     val TransactionTimestamp        = FunctionDef[Any, ZonedDateTime](FunctionName("transaction_timestamp"))
+    val Encode                      = FunctionDef[(String, String), String](FunctionName("encode"))
+    val Decode                      = FunctionDef[(String, String), String](FunctionName("decode"))
```
jczuchnowski, line 53: Encode and decode in PostgreSQL operate on binary string type (`bytea`). The documentation describes the difference https://www.postgresql.org/docs/13/datatype-binary.html. I'm wondering if it wouldn't be more appropriate to represent these as `Chunk[Byte]` and introduce tests that include non-printable octets.

danieletorelli: I'm definitely with you about it and I had the same doubt but I eventually, opted to use `String` since also `md5`, that is supposed to ingest a `bytea`, was defined as:
```
val Md5                         = FunctionDef[String, String](FunctionName("md5"))
```
probably we should also change this one then. WDYT?

I'll try to adapt them all to `Chunk[Byte]`.

jczuchnowski: Right, apparently I missed that. So yeah, once we have `encode`/`decode` working we can fix others.

danieletorelli: I pushed my proposal. I had to make invariant `zio.sql.TypeTagModule.TypeTag` and other two as a consequence in order to avoid this compilation error when mapping over the chunk:
```
.../zio-sql/postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala:120:64
value map is not a member of Any
        case tt @ TByteArray      => render(tt.cast(lit.value).map("""\%03o""" format _).mkString("E\'", "", "\'"))
```

Please suggest alternatives if you think this way is not right.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -9,32 +9,32 @@ trait TypeTagModule {
 
   type TypeTagExtension[+A]
 
-  sealed trait TypeTag[+A] {
+  sealed trait TypeTag[A] {
```
robmwalsh, line 12: the covariance makes things much safer RE casting (i.e. you can't do it incorrectly using `cast`) so I think this should stay.

core/jvm/src/main/scala/zio/sql/typetag.scala
```diff
@@ -9,32 +9,32 @@ trait TypeTagModule {
...
-    implicit case object TLocalTime                                               extends NotNull[LocalTime]
-    implicit case object TLong                                                    extends NotNull[Long]
-    implicit case object TOffsetDateTime                                          extends NotNull[OffsetDateTime]
-    implicit case object TOffsetTime                                              extends NotNull[OffsetTime]
-    implicit case object TShort                                                   extends NotNull[Short]
-    implicit case object TString                                                  extends NotNull[String]
-    implicit case object TUUID                                                    extends NotNull[UUID]
-    implicit case object TZonedDateTime                                           extends NotNull[ZonedDateTime]
-    sealed case class TDialectSpecific[+A](typeTagExtension: TypeTagExtension[A]) extends NotNull[A]
+    sealed trait NotNull[A]                                                      extends TypeTag[A]
```
robmwalsh, line 17: as above

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -115,10 +117,8 @@ trait PostgresModule extends Jdbc { self =>
     private[zio] def renderLit[A, B](lit: self.Expr.Literal[_])(implicit render: Renderer): Unit = {
       import TypeTag._
       lit.typeTag match {
-        case tt @ TByteArray      => render(tt.cast(lit.value))                     // todo still broken
-        //something like? render(tt.cast(lit.value).map("\\\\%03o" format _).mkString("E\'", "", "\'"))
-        case tt @ TChar           =>
-          render("'", tt.cast(lit.value), "'") //todo is this the same as a string? fix escaping
+        case tt @ TByteArray      => render(tt.cast(lit.value).map("""\%03o""" format _).mkString("E\'", "", "\'"))
```
robmwalsh, line 120: let's just cast here where it's not inferred correctly so we can keep covariance:
```scala
case TByteArray => render(lit.value.asInstanceOf[Chunk[Byte]].map("""\%03o""" format _).mkString("E\'", "", "\'")) //todo fix `cast` infers correctly but map doesn't work for some reason
```

robmwalsh (changes requested):
@danieletorelli looks good. I'd prefer to keep the covariance as it makes things much safer. It looks like Chunk[Byte] is inferred correctly but map isn't working for some reason. I think manually casting is sufficient for now and we can investigate why `map` doesn't work here later.

danieletorelli:
> @danieletorelli looks good. I'd prefer to keep the covariance as it makes things much safer. It looks like Chunk[Byte] is inferred correctly but map isn't working for some reason. I think manually casting is sufficient for now and we can investigate why `map` doesn't work here later.

@robmwalsh Sure, makes sense. I readapted the code to your proposed change.

jczuchnowski (approved):

jczuchnowski (approved):

PR #361, opened 2021-01-08
Issue/328

maciejbak85:
By doing change, only 1 DB container will be executed for one suite.
It speeds up tests a lot, and also consumes less memory.


jczuchnowski:
Awesome work @maciejbak85 !

What about the case we'd like to run a container per test after all? It might be needed when testing inserts and updates. Is there a way to do this now?

mysql/src/test/scala/zio/sql/mysql/FunctionDefSpec.scala
```diff
@@ -9,7 +9,7 @@ object FunctionDefSpec extends MysqlRunnableSpec with ShopSchema {
   import Customers._
   import FunctionDef._
 
-  val spec = suite("Mysql FunctionDef")(
+  override def specLayered = suite("Mysql FunctionDef")(
```
jczuchnowski, line 12: Why this change?

maciejbak85: to avoid duplicating addCustomLayerShared at the end ;) its not a big deal but well I think it always adds some value.

maciejbak85:
> Awesome work @maciejbak85 !

thanks!

> 
> What about the case we'd like to run a container per test after all? It might be needed when testing inserts and updates. Is there a way to do this now?

for now its not available, I did not want to keep unused class in the code. Do you think its ok to keep it ? If yes then I can add new trait.

robmwalsh:
Thanks @maciejbak85, really looking forward to not having so many containers running :D this will also mitigate the impact of leaking test containers because there aren't so many around.

I think this behaviour would be best controlled by creating a new `TestAspect` which runs the enclosed tests in a new container. Then we can do something like

`@@ newContainer`

on individual tests or entire suites that should run in a new container

we can add it to `def aspects` in JDBCRunnableSpec to apply newContainer aspect at the spec level, then we don't have explain the `specLayered` thing to everyone while achieving the best of both worlds.

jczuchnowski:
@robmwalsh this is a great idea. I think in the end we could end up with a separate zio-testcontainers library for zio-test integration.

That said, I'll merge this PR and create a separate issue for it. The current solution will already unblock us on multiple failing test cases.

jczuchnowski (approved):

jczuchnowski:
Congratulations on your first ZIO SQL contribution @maciejbak85 :beer:

maciejbak85:
hmm interesting idea, I will try to find some time to look into that solution :)

PR #406, opened 2021-03-04
Hackday cleanup

jdegoes:
Miscellaneous cleanup coming from ZIO SQL hack day.

Needs further tests and fixing whatever's broken, but it's in a good position for handoff!

/cc @jczuchnowski


jczuchnowski:
There are no obvious problems with this PR. The only problem was formatting. So the only thing that remains is adding tests for `ZTransaction` and `ConnectionPool`. I'll deal with it this week.

Issue #184, opened 2020-11-14
Add 'format' function to PostgresModule


jczuchnowski:
Add support for function `format` described in the PostgreSQL [docs](https://www.postgresql.org/docs/current/functions-string.html).
1. Add function to `PostgresFunctionDef` in `PostgresModule`
2. Add test in `FunctionDefSpec`

angelicarao:
I'd like to work on this one

TobiasPfeifer:
working on this

PR #415, opened 2021-03-28
Add 'format' function to PostgresModule #184

TobiasPfeifer:
implements format function in ProstgresModule #184


postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
```diff
@@ -227,6 +227,12 @@ trait PostgresModule extends Jdbc { self =>
       FunctionDef[Timestampz, Timestampz](FunctionName("make_timestamptz"))
     val Encode                      = FunctionDef[(Chunk[Byte], String), String](FunctionName("encode"))
     val Decode                      = FunctionDef[(String, String), Chunk[Byte]](FunctionName("decode"))
+    val Format0                      = FunctionDef[String, String](FunctionName("format")) // TODO: varargs
```
TobiasPfeifer, line 230: varargs not available yet (see also 'concat' function)
Is there an open issue for varargs support? otherwise I would open an issue to track this (and might give it a try)

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -151,7 +151,105 @@ object FunctionDefSpec extends PostgresRunnableSpec with ShopSchema {
...
+            s"""Person: Jose Wiggins with double quoted "identi fier" """
+          )
+
+          val testResult = execute(query.to[String, String](identity))
+          collectAndCompare(expected, testResult)
+        },
+        testM("format4") {
+          import Expr._
+
+          val query = select(Format4("Person: %s %s with null-literal %L and non-null-literal %L ", Customers.fName, Customers.lName, "FIXME: NULL", "literal")) from customers
```
TobiasPfeifer, line 223: couldn't find a solution to specify NULL
tried `Option.empty[String]` and `null.asInstanceOf[String]`

postgres/src/main/scala/zio/sql/postgresql/PostgresModule.scala
jczuchnowski: Created issue here https://github.com/zio/zio-sql/issues/439

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
jczuchnowski: Yeah, Options are not well supported at this point.

jczuchnowski:
Thanks for this @TobiasPfeifer !

jczuchnowski (approved):

PR #427, opened 2021-04-09
Support select statements without `from`

### LINKED ISSUE #323: Support `select` statements without `from`
At the moment we need to select from a table, even if we have no requirement for a table
e.g. we can't execute `select 1`, but we can execute `select 1 from aTable`

Existing tests should be cleaned up to remove tables from queries where none are required, which should be proof enough that this works.

TobiasPfeifer:
fixes #323


jczuchnowski (approved):

jczuchnowski:
@TobiasPfeifer awesome work here, thanks!

PR #446, opened 2021-05-04
Remove curly braces from examples dsl

jczuchnowski:


robmwalsh:
What's wrong with the curly braces? I personally find it easier to read with them

jczuchnowski:
The main problem I see, is that it's irregular and I want the examples to be presented in the simplest form possible.
That's also the reason I've changed joins from the infix notation to normal method call. Otherwise some of the parenthesis position look kind of confusing.

The second reason is that I'm trying to get everything to compile in Scala 3 and there are a lot of errors right now, some of them with our DSL, and I just want to make sure we're not hitting any edge cases here.

robmwalsh (approved):

Issue #218, opened 2020-11-16
Add `crossApply` and `outerApply` to SQL Server Module


robmwalsh:

sviezypan:
I'll work on this

sviezypan:
Done in #529

PR #475, opened 2021-06-28
#218 cross apply & outer apply initial implementation

sviezypan:
Hello @jczuchnowski @robmwalsh 
I wrote some code and some TODOs for this task and would like to get some feedback if I am going in the right direction.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -207,10 +207,12 @@ lazy val sqlserver = project
   .settings(buildInfoSettings("zio.sql.sqlserver"))
   .settings(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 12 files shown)


robmwalsh:
Hey @sviezypan, thanks a lot for your work on this! I should be able to look at this in more detail later this week, but as some quick initial feedback, to my knowledge, cross apply and outer apply are only supported by SQL server, so ideally we should keep everything related to it in the SQL server subproject. If you haven't heard more from me by the weekend, please prod me, because I've probably forgotten ;)

sviezypan:
I will open a new PR for this issue.

PR #477, opened 2021-07-01
rederDelete for SqlServer

eqk:


mysql/src/test/scala/zio/sql/TestContainer.scala
```diff
@@ -16,11 +16,11 @@ object TestContainer {
       }
     }(container => effectBlocking(container.stop()).orDie).toLayer
 
-  def mysql(imageName: String = "mysql"): ZLayer[Blocking, Throwable, Has[MySQLContainer]] =
+  def mysql(imageName: String = "mariadb"): ZLayer[Blocking, Throwable, Has[MySQLContainer]] =
     ZManaged.make {
       effectBlocking {
         val c = new MySQLContainer(
-          mysqlImageVersion = Option(imageName).map(DockerImageName.parse)
+          mysqlImageVersion = Option(imageName).map(DockerImageName.parse).map(_.asCompatibleSubstituteFor("mysql"))
```
jczuchnowski, line 23: This shouldn't be part of this PR.

jczuchnowski (changes requested):
Great work! But in addition to that, could you please add a new ReadSpec with two short tests that would check for a proper handling of `.isTrue` and `.isNotTrue` ? It's very specific to SQL Server and I wouldn't want it to fall through the cracks at some point.

eqk:
> Great work! But in addition to that, could you please add a new ReadSpec with two short tests that would check for a proper handling of `.isTrue` and `.isNotTrue` ? It's very specific to SQL Server and I wouldn't want it to fall through the cracks at some point.

I have some problems with ReadSpec, because of default where clause in select builder [here](https://github.com/zio/zio-sql/blob/master/core/jvm/src/main/scala/zio/sql/select.scala#L13) . With initial query like this 
`val query =
  select(Count(customerId))
    .from(customers)
    .where(verified.isTrue)`

It renders to this 
`select count(customers.id) from customers where true and customers.verified = 1`
Witch is not valid. What is the purpose of this clause? Maybe redefine SelectBuilder in this module? @jczuchnowski

eqk:
#480 need to redefine `where true` interpretation in SQLServer module

robmwalsh:
How about changing to an Option for the where expression in Read.Select?

sviezypan:
hi @eqk problems with sql server and booleans were resolved in #529 , if you would resolve conflicts on this PR, we could merge it

sviezypan (approved):

jczuchnowski (approved):

PR #600, opened 2022-02-23
Bug fixes regarding `Having` and column types 

sviezypan:
@jczuchnowski @jdegoes 
PR contains two bugfixes
1. `having` is much more sound. It can only be called:

-  If its called with an aggregated function returning boolean like `Having Count(id) > 5`, while all the previously selected columns appeared in group by clause.
- If its called with a normal expression returning boolean like `having customer_id = '636ae137-5b1a-4c8c-b11f-c47c624d9cdc` and all the previously selected columns appeared in group by clause.
Examples of what compiles and what does not are in `GroupByHavingSpec.scala`

 
2. columns of same name from different table are treated as different columns. Phantom type F now contains also reference to table type `Features.Source["column name".type, tableType]` As a result we are able to make more complex queries sound. In the following query `city.name` and `ms.name` are different columns and compiler requires us to group by both of them. Test is in `PostgresModuleSpec.scala`
```
select(metroSystemName ++ cityName ++ Count(metroLineId)())
        .from(
          metroLine
            .join(metroSystem)
            .on(metroSystemId === systemId)
            .join(city)
            .on(cityIdFk === cityId)
        )
        .groupBy(metroSystemName, cityName)
```


core/jvm/src/test/scala/zio/sql/GroupByHavingSpec.scala
```diff
@@ -37,9 +37,72 @@ object AggregatedProductSchema {
...
+
+  select(Sum(price))
+    .from(productTable)
+    .groupBy(name)
+    .having(name > 10)
+
+  select(price)
+    .from(productTable)
+    .groupBy(price)
+    .having(Count(price) > 10)
```
jczuchnowski, line 86: This isn't really a spec. Maybe move it to examples?

jczuchnowski (approved):

jczuchnowski (approved):

Issue #604, opened 2022-03-15
Connections are not released if acquire is interrupted


petoalbert:
[ConnectionPoolLive.connection](https://github.com/zio/zio-sql/blob/edcae6042abb4ebe7ee8cd323572ac9c7cbf9b18/jdbc/src/main/scala/zio/sql/ConnectionPool.scala#L96) returns a Managed instance that is responsible for releasing the connection after usage. If no connection is instantly available, a new `Promise[Nothing, ResettableConnection]` will be created and put into a queue, and the `acquire` step will be completed once the promise is completed. However, if the fiber running the acquire phase is interrupted, nothing will handle the connection that the promise completes with, and specifically **that connection will never be released**. The `onInterrupt` block only does a best-effort job at releasing the connection on interrupt if it is already available. 

To demonstrate this issue, I have prepared the following snippet. Apologies for the artificial example using `Transaction.fromEffect`, but this is the simplest way to demonstrate it. We can imagine that the transactions in this example represent very slow queries.

```scala
// hold the connection until the promise is completed
def slowTransaction(promise: Promise[Nothing, Unit]) = 
    execute(ZTransaction.fromEffect(promise.await)).useNow
    
// time out after 1 second. This will interrupt the fiber while the connection hasn't been acquired
val timeoutTransaction =
    execute(ZTransaction.fromEffect(ZIO.succeed(()))).timeout(1.second).useNow
    
val simpleTransaction =
    execute(ZTransaction.fromEffect(ZIO.succeed(()))).useNow


for {
    p <- Promise.make[Nothing, Unit]
    // acquire all connections from the pool
    _ <- ZIO.replicateZIO(poolSize)(slowTransaction(p).fork)
    // wait for all connections from the pool
    _ <- ZIO.replicateZIO(poolSize)(timeoutTransaction.fork)
    // wait with the "slow queries" to let everyone else time out
    _ <- ZIO.sleep(2.second)
    _ <- p.succeed(())
    // simpleTransaction will wait forever for a connection!!!
    // the connection pool is essentially dead by this time
    _ <- simpleTransaction
} yield ()
```

I have thought about multiple ways to solve this and I have prepared a fix for this, I would be happy to discuss it with you.

PR #605, opened 2022-03-15
Ensure that connections are released after interrupt

petoalbert:
#604


sviezypan:
hey @petoalbert thanks so much for your work on this. This is a great catch. 
Could you add example from #604 as a test? 
Also please run `sbt fmt` to format the code.

sviezypan (approved):

PR #618, opened 2022-04-02
Improve inserting some & none fixes #585

### LINKED ISSUE #585: improve on inserting optional values
Let's have a table with nullable column:
```

val persons = ((string("name") ++ (int("age") @@ nullable)).table("person")
val (name , age) = persons.columns
```
result type of `age` is then `Expr[F, T, Option[Int]]`

When inserting user defined case class its easy to insert `None`

```
case class Person(name: String, age: Option[Int])
implicit val schema = DeriveSchema.gen[Person]

val data = List(Person("my_name", None), Person("other_name", Some(42)))

insertInto(persons)(name ++ age).values(data)
```

however in case we are inserting tuples and we want to insert Some or None to `age` we need to cast it to `Option[LocalDate]`

```
val data = List(("my_name", Option.empty[Int])), ("other_name", Option(42)))

insertInto(persons)(name ++ age).values(data)
```

This is because of strict type checks that accompany insert module. ZIO schema for tuple is derived automatically - in this case Schema[(String, Option[Int])] - but based on inserted values, compiler search for Schema[(String, None)] or Schema[(String, Some[Int])].

So we need to play around with inserting tuples more, ideally we want to be able to do:
```
val persons = List(("my_name", None), ("other_name", Some(42)))

insertInto(persons)(name ++ age).values(persons)
```

sviezypan:
Old way
```
insertInto(persons)(personId ++ fName ++ lName ++ dob)
        .values((UUID.randomUUID(), "Charles", "Dent", Option(LocalDate.of(2022, 1, 31))))

insertInto(persons)(personId ++ fName ++ lName ++ dob)
        .values((UUID.randomUUID(), "Martin", "Harvey", Option.empty[LocalDate]))
```

new way

```
insertInto(persons)(personId ++ fName ++ lName ++ dob)
        .values((UUID.randomUUID(), "Charles", "Dent", Some(LocalDate.of(2022, 1, 31))))

insertInto(persons)(personId ++ fName ++ lName ++ dob)
        .values((UUID.randomUUID(), "Martin", "Harvey", None))
```

Plus two tiny things, bug fixes:
- I removed Zmanaged/Scope from environment when transaction is executed. 
- transaction is rollbacked by any defect or error.
- changed mssql/server to azure-sql-edge container for compatibility with arm based machines

### DIFF
--- a/jdbc/src/main/scala/zio/sql/ConnectionPool.scala
+++ b/jdbc/src/main/scala/zio/sql/ConnectionPool.scala
@@ -15,7 +15,7 @@ trait ConnectionPool {
    * The managed resource will safely acquire and release the connection, and
    * may be interrupted or timed out if necessary.
    */
-  def connection: Managed[Exception, Connection]
+  def connection: ZIO[Scope, Exception, Connection]
 }
 object ConnectionPool {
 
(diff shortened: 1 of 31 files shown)


jdbc/src/main/scala/zio/sql/ConnectionPool.scala
```diff
@@ -11,7 +11,7 @@ import zio.sql.ConnectionPool.QueueItem
 trait ConnectionPool {
 
   /**
-   * Retrieves a JDBC java.sql.Connection as a [[zio.Managed]] resource.
+   * Retrieves a JDBC java.sql.Connection as a [[ZIO[Scope, Exception, Connection]]] resource.
```
jczuchnowski, line 14: Good catch :smile:

jczuchnowski (dismissed):

jczuchnowski (dismissed):

jczuchnowski (approved):

PR #620, opened 2022-04-04
Upgrade To ZIO 2.0.0-RC4

adamgfraser:


jdbc/src/main/scala/zio/sql/SqlDriverLiveModule.scala
```diff
@@ -45,7 +45,7 @@ trait SqlDriverLiveModule { self: Jdbc =>
 
     def read[A](read: Read[A]): Stream[Exception, A] =
       ZStream
-        .managed(pool.connection)
+        .scoped(pool.connection)
         .flatMap(readOn(read, _))
```
sviezypan, line 49: Don't we need Stream.scoped() alternative that would accept the ZStream?
Just wondering what's the difference between
```
ZStream
        .scoped(pool.connection)
        .flatMap(readOn(read, _))
```
and 
```
ZStream
        .scoped(pool.connection.flatMap(readOn(read, _)))
```
Second alternative is currently not possible.

sviezypan (changes requested):
hey @adamgfraser thank you very much for this. I am actually struggling for a while with upgrading to RC3 and RC4. I arrived at the more or less same solution at my branch but I am having problem with running tests that is also visible here.

This is build from different PR from Saturday https://github.com/zio/zio-sql/runs/5799709220?check_suite_focus=true
there are also tests from Postgresql, Mysql and MsSqlServer module.

After upgrade to RC4 only tests from core module are launched. You can see it in the build of this PR https://github.com/zio/zio-sql/runs/5819972273?check_suite_focus=true I also tried to use zio-managed dependency (so only change was version increase and couple of imports) and the result was the same.

jdbc/src/main/scala/zio/sql/SqlDriverLiveModule.scala
adamgfraser: There is `ZStream.unwrapScoped` which is equivalent to the previous `ZStream.unwrapManaged` and those were the only two operators for converting a `ZManaged` into a `ZStream` so I don't think there is any missing operator. If I understand the signatures correctly `pool.connection` is a resource so the second wouldn't compile because `connection is a `ZManaged` or `ZIO` and `readOn` returns a stream.

adamgfraser:
@sviezypan That's great feedback! I think the issue is most likely to be in the test framework rather than the resource management itself. We moved to a new streaming format for test output in `RC4` which required a significant redesign of the internals of ZIO Test. We've made some changes that are relevant to this so I would try with the latest snapshot first and then if you are still having issues we can spend more time on it.

Copying @swoogles.

adamgfraser:
@sviezypan It looks like all tests are passing now for real this time. Would you please take a look when you have the chance?

sviezypan (approved):
Looks great. Thank you so much @adamgfraser.

Issue #584, opened 2022-02-14
support alternative selection of columns


sviezypan:
In order to resemble SQL as much as possible, I think it would be nice to support also selections of columns separated by commas in addition to supporting selection with ++ operator.

```
// curent 
select(name ++ age).from(persons)
insertInto(persons)(name ++ age).values(...)

// additional alternative
select(name, age).from(persons)
insertInto(persons)(name, age).values(...)
```

sviezypan:
done in #630

PR #630, opened 2022-04-19
Alternative selection of columns #584

sviezypan:
Old way (still supported)
```
insertInto(customers)(
        customerId ++ dob ++ fName ++ lName ++ verified ++ created
      ).values(data)
      
select(id ++ lastUpdated ++ baseAmount ++ deleted) from productTable
```
new way
```
insertInto(customers)(
        customerId, dob, fName, lName, verified, created
      ).values(data)
      
select(id, lastUpdated, baseAmount, deleted) from productTable
```


jdegoes (dismissed):

jczuchnowski:
This looks great, but I think we should still have the old way on equal rights in the documentation. The ++ operator makes the selection sets composable - something that the comma DSL doesn't allow I suppose.

jczuchnowski (approved):

PR #632, opened 2022-04-24
Render quoted table names and column names for PostgreSQL

### LINKED ISSUE #142: Render all table and column names in double quotes in PostgreSQL
Modify `PostgresModule#renderRead` so that the queries it renders are case sensitive. 

For example:
```sql
select users.first_name, users.last_name from users
```
should become
```sql
select "users"."first_name", "users"."last_name" from "users"
```

We will also need appropriate tests in `PostgresModuleTest` that will prove uppercased db objects can be correctly queried from ZIO SQL.

jczuchnowski:
Closes #142


postgres/src/test/scala/zio/sql/postgresql/PostgresModuleSpec.scala
```diff
@@ -691,9 +691,7 @@ object PostgresModuleSpec extends PostgresRunnableSpec with DbSchema {
       assertM(result)(equalTo(expected))
     },
     test("update rows") {
-      import Persons._
-
-      val query = update(persons).set(fName, "Charlie").where(fName === "Charles")
+      val query = update(customers).set(fName, "Jaroslav").where(fName === "Jaro")
```
sviezypan, line 694: Will I be famous now?

sviezypan (approved):

postgres/src/test/scala/zio/sql/postgresql/PostgresModuleSpec.scala
jczuchnowski: All the bots will know you.

PR #663, opened 2022-05-24
Feature645:  Batch processing ( insert, update, and delete)

balanka:


build.sbt
```diff
@@ -26,6 +26,7 @@ addCommandAlias("check", "all scalafmtSbtCheck scalafmtCheck test:scalafmtCheck"
 
 val zioVersion                 = "2.0.0-RC6"
 val zioSchemaVersion           = "0.1.9"
+//val zioSchemaVersion           = "0.2.0-RC5"
```
sviezypan, line 29: remove comment

jdbc/src/main/scala/zio/sql/TransactionModule.scala
```diff
@@ -87,15 +87,28 @@ trait TransactionModule { self: Jdbc =>
...
+    def apply[Z: Schema](insert: List[self.Insert[_, Z]]): ZTransaction[Any, Exception, List[Int]] =
+      txn.flatMap { case Txn(connection, coreDriver) =>
+        ZTransaction.fromEffect(coreDriver.insertOnBatch(insert, connection))
+      }
+
     def apply(delete: self.Delete[_]): ZTransaction[Any, Exception, Int] =
       txn.flatMap { case Txn(connection, coreDriver) =>
         ZTransaction.fromEffect(coreDriver.deleteOn(delete, connection))
       }
+    def applyBatch(delete: List[self.Delete[_]]): ZTransaction[Any, Exception, List[Int]] =
```
sviezypan, line 108: I would just call it apply, not applyBatch.

jdbc/src/main/scala/zio/sql/jdbc.scala
```diff
@@ -32,9 +38,18 @@ trait Jdbc extends zio.sql.Sql with TransactionModule with JdbcInternalModule wi
   def execute(delete: Delete[_]): ZIO[SqlDriver, Exception, Int] =
     ZIO.serviceWithZIO(_.delete(delete))
 
+  def executeBatch(delete: List[Delete[_]]): ZIO[SqlDriver, Exception, List[Int]] =
```
sviezypan, line 41: Lets keep it consistent and call all the `batch` methods executeBatch, not only the one for delete

postgres/src/test/scala/zio/sql/postgresql/DeleteBatchSpec.scala
```diff
@@ -0,0 +1,36 @@
+package zio.sql.postgresql
+
+import zio.Cause
+import zio.test.Assertion._
+import zio.test._
+
+object DeleteBatchSpec extends PostgresRunnableSpec with DbSchema {
```
sviezypan, line 7: Could you please add test where you would execute multiple queries and then sum the resulting number of changed lines? Also one test for insert and update would be nice.

sviezypan (changes requested):
hi @balanka thank you very much for your work on this. Overall this looks nice! Couple of minor comments, also you probably need to format.

jdbc/src/main/scala/zio/sql/TransactionModule.scala
balanka: ok

jdbc/src/main/scala/zio/sql/jdbc.scala
balanka: You are right. Actually i wanted to name them all uniformly `executeBatch` 
The problem is  that does not compile.
The compiler complains about `double definition
def executeBatch(delete: List[Jdbc.this.Delete[_]]): zio.ZIO[Jdbc.this.SqlDriver,Exception,List[Int]] at line 41 and
[error] def executeBatch(update: List[Jdbc.this.Update[_]]): zio.ZIO[Jdbc.this.SqlDriver,Exception,List[Int]] at line 53
[error] have same type after erasure: (delete: List): zio.ZIO
` 
therefore  i chose different name  `executeBatchDelete, executeBatchInsert and executeBatchUpdate`

build.sbt
balanka: done

jdbc/src/main/scala/zio/sql/TransactionModule.scala
balanka: the issue here is  i cannot use the same apply method for update and delete
as these method names would collide after type erasure
```
 double definition:
[error] def applyBatch(update: List[TransactionModule.this.Update[_]]): TransactionModule.this.ZTransaction[Any,Exception,List[Int]] at line 90 an
[error] def applyBatch(delete: List[TransactionModule.this.Delete[_]]): TransactionModule.this.ZTransaction[Any,Exception,List[Int]] at line 108
[error] have same type after erasure: (update: List): zio.sql.TransactionModule#ZTransaction
[error]     def applyBatch(delete: List[self.Delete[_]]): ZTransaction[Any, Exception, List[Int]] =
```

postgres/src/test/scala/zio/sql/postgresql/DeleteBatchSpec.scala
balanka: done

jdbc/src/main/scala/zio/sql/TransactionModule.scala
sviezypan: I see but the names are really sub optimal. From the user perspective, its hard to discover these methods.

I cannot think of much better names but probably these are better:
- ZTransaction.batchUpdate
- ZTransaction.batchDelete
- ZTransaction.batchInsert

What do you think?

sviezypan (changes requested):
One minor suggestion for names. Seems that you need to format the code and tests failed for 2.12, can you check? Otherwise good to go.

jdbc/src/main/scala/zio/sql/TransactionModule.scala
balanka: fine.
Let me rename the apply method accordingly

build.sbt
balanka: done

sviezypan (approved):

sviezypan:
@balanka you still need to sign CLA. Click here https://cla-assistant.io/zio/zio-sql?pullRequest=663 scroll down and click on `Sign in with Github to agree`

balanka:
Due to some trouble related to the signing of the CLA I need to close and reopen this PR

Issue #597, opened 2022-02-20
Cleanup the 'examples' project


jczuchnowski:
Currently the examples are spread around the two files with names Examples and Exmaple1. 
We need to review the examples and merge them into one consistent file.

sushengloong:
👋 hi, I would like to work on this ticket.

PR #668, opened 2022-05-26
Clean up examples

sushengloong:
- To fix issue https://github.com/zio/zio-sql/issues/597
- Add missing examples to Examples.scala and remove Examples1.scala.
- Change the example from using `SqlServerJdbcModule` to `PostgresJdbcModule`. Aside from supporting Open Source DBMS, certain functions don't work with `SqlServerJdbcModule`, eg. `Offset`. That said, I also found out that `As` doesn't work for Postgres and MySQL but it's out of scope for this PR.

Feedback is very much appreciated!


jczuchnowski:
I didn't expect switching to PostgreSQL but considering it's our most advanced module it totally makes sense :+1:

jczuchnowski (approved):

Issue #653, opened 2022-05-23
Add `current_time` function to MysqlSqlModule


jczuchnowski:
Add support for function `current_time` described in the MySQL [docs](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_current-time).

Add function to MysqlFunctionDef in MysqlSqlModule
Add test in FunctionDefSpec

amrkamel:
I would like to work on this issue

PR #669, opened 2022-05-26
[zio-hackathon-2022] Add current_time function to MysqlSqlModule

amrkamel:
Resolves https://github.com/zio/zio-sql/issues/653

PR #681, opened 2022-05-26
Add current_time function to MysqlSqlModule

amrkamel:
Resolves https://github.com/zio/zio-sql/issues/653

Second attempt after technical difficulties in previous PR https://github.com/zio/zio-sql/pull/669


jczuchnowski (approved):

mysql/src/main/scala/zio/sql/mysql/MysqlSqlModule.scala
```diff
@@ -1,7 +1,7 @@
 package zio.sql.mysql
 
-import java.time.{ LocalDate, LocalTime, Year, ZonedDateTime }
```
jczuchnowski, line 3: LocalTime needs to stay

jczuchnowski (approved):

Issue #158, opened 2020-11-12
Implement renderInsert for MySQL


jczuchnowski:

sushengloong:
I'd like to work on this one

PR #685, opened 2022-05-26
Implement renderInsert for MySQL

sushengloong:
- Address https://github.com/zio/zio-sql/issues/158
- Mostly adapted from the Postgres implementation.

### DIFF
--- a/mysql/src/main/scala/zio/sql/mysql/MysqlRenderModule.scala
+++ b/mysql/src/main/scala/zio/sql/mysql/MysqlRenderModule.scala
@@ -1,7 +1,11 @@
 package zio.sql.mysql
 
+import java.time._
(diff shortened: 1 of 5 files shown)


sviezypan:
@sushengloong thank you for your work on this! Can you please add one more test that would insert tuples containing some optional and also localDateTime ?

sviezypan:
@sushengloong You need to add implicit schema of LocalDate to the module
Something like
```
  implicit val localDateSchema =
    Schema.primitive[LocalDate](zio.schema.StandardType.LocalDateType(DateTimeFormatter.ISO_DATE))
 ```
 but please make sure that formatting of date is correct - I mean from MySQL perspective

sviezypan (approved):
Looks great! Ready to merge it when CI passes.

PR #692, opened 2022-05-27
Add SqlTransaction type

### LINKED ISSUE #665: Adopt ZIO JDBC model for transactions
**ISSUE**

ZIO SQL was created before ZIO 2. As a result, it utilizes an older pattern for transactions, which involves the `ZTransaction` type. This means that users must be painfully aware of transaction boundaries and transaction types, and non-transaction types must be lifted up into `ZTransaction`.

**SOLUTION**

With layers as eliminators, ZIO JDBC has pioneered a new way to think about transactions, foreshadowing the work done in ZIO 2 which replaces `ZManaged` with scoped ZIO effects. These new ways of dealing with transactions and resources will permit a much cleaner way to model transactions in ZIO SQL. In particular, transactions can become ordinary `ZIO` effects, but with an environment parameter that will be eliminated by a `transaction` operator.

Possibly the end-user API can look like this:

```scala
val myQuery : ZIO[SqlTransaction, Throwable, Result] = ???

transaction { myQuery } // ZIO[Any, Throwable, Result]
```

**IMPACT**

Making this change will bring ZIO SQL up to the latest standards for dealing with resources and transactions in the ZIO 2 ecosystem, and make use of the library by end-users much easier. 

**WHERE TO LOOK**

Study the design of ZIO JDBC and ZIO 2 scopes. Then search the code base for `ZTransaction`, which will need to be replaced by an ordinary `ZIO` effect, where the environment is used to keep track of transaction information.

**PROBLEMMATIC ISSUES**

It may be necessary to understand scopes and to take a peek at ZIO JDBC in order to complete this issue successfully. Reach out to maintainers for hints or pairing.

mschuwalow:
resolves #665


jdbc/src/main/scala/zio/sql/ExprSyntaxModule.scala
```diff
@@ -0,0 +1,27 @@
+package zio.sql
+
+import zio._
+import zio.stream.ZStream
+import zio.schema.Schema
+
+trait ExprSyntaxModule { self: Jdbc =>
+  implicit final class ReadSyntax[A](self: Read[A]) {
```
mschuwalow, line 8: imo this end up being the best end-user syntax.

jdbc/src/main/scala/zio/sql/jdbc.scala
```diff
@@ -37,4 +44,7 @@ trait Jdbc extends zio.sql.Sql with TransactionModule with JdbcInternalModule wi
 
   def execute(update: Update[_]): ZIO[SqlDriver, Exception, Int] =
```
mschuwalow, line 45: We may want to rename this to make clearer that it does not run as part of a transaction

jdbc/src/main/scala/zio/sql/ExprSyntaxModule.scala
sviezypan: what about `toZIO` ? 
imho It seems weird from user perspective to have to call both `transact` and `run`.
Ideally if we could omit calling `run` somehow....how ZIO JDBC does that?

But overall this is great simplification, I'll let also @jczuchnowski review

mschuwalow: I'm not sure.

The semantics are that you will run the individual statements when you use run on them, but only commit them after the transact scope is done.
I think we should big naming that makes this very clear.

Maybe `withTransaction` instead of `transact`?

jczuchnowski (approved):

PR #693, opened 2022-05-27
insert examples for tuple and case class with scema derive

### LINKED ISSUE #672: Add examples for inserts
In the module `examples` add some for insert operation.
Should be enough to include one example for inserting tuples and one for inserting case class (using ZIO Schema).
Explanation of these can be found here https://zio.github.io/zio-sql/docs/overview/deep_dive#insert-tuples

gantu:
resolves #672


sviezypan (approved):
Thank you!

Issue #156, opened 2020-11-12
Implement renderDelete for MySQL


jczuchnowski:
Following the pattern from `renderRead` in PostgresModule and SqlServerModule implement a MySQL specific `renderDelete`.

jczuchnowski:
Done in https://github.com/zio/zio-sql/pull/424

PR #694, opened 2022-05-27
Render delete for oracle

bburdiliak:
implements #156 

I had to specially deal with boolean values, as Oracle does support them only via smallint (or char).

### DIFF
--- a/oracle/src/test/scala/zio/sql/oracle/OracleModuleSpec.scala
+++ b/oracle/src/test/scala/zio/sql/oracle/OracleModuleSpec.scala
@@ -10,43 +10,13 @@ object OracleModuleSpec extends OracleRunnableSpec with ShopSchema {
 
   import Customers._
 
-//  private def customerSelectJoseAssertion[F: Features.IsNotAggregated](
-//    condition: Expr[F, customers.TableType, Boolean]
-//  ) = {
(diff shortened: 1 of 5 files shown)


jczuchnowski:
> I had to specially deal with boolean values, as Oracle does support them only via smallint (or char).
Yeah, same problem with SQL Server.

This looks good. Thank you!

jczuchnowski (approved):

oracle/src/test/scala/zio/sql/oracle/OracleModuleSpec.scala
```diff
@@ -0,0 +1,33 @@
+package zio.sql.oracle
+
+import zio.test.Assertion._
+import zio.test.TestAspect._
+import zio.test._
+
+import scala.language.postfixOps
+
+object OracleModuleSpec extends OracleRunnableSpec with ShopSchema {
```
baldram, line 9: The tested module seems to be called `OracleSqlModule` so the test would be `OracleSqlModuleSpec`.

baldram: It's merged so I will rename it within my change.

jczuchnowski: :+1:

PR #695, opened 2022-05-27
Implement renderInsert for Oracle

peixunzhang:


peixunzhang:
Hi, I don't if I should get rid of the duplications between the Postgres module and this one.

sviezypan (changes requested):
Thanks for you work. Can you please also add few tests, something like in #685  ?

sviezypan:
@peixunzhang seems that some tests have problems, can you please check?

oracle/src/test/scala/zio/sql/oracle/OracleSqlModuleSpec.scala
```diff
@@ -57,6 +64,195 @@ object OracleSqlModuleSpec extends OracleRunnableSpec with ShopSchema {
...
+          offsettimeCol,
+          shortCol,
+          stringCol,
+          uuidCol,
+          zonedDatetimeCol,
+          yearMonthCol,
+          durationCol
+        ).values(row)
+
+        // TODO: ensure we can read values back correctly
```
sviezypan, line 219: has this worked? test like this would be awesome!

sviezypan (commented):
Looks good, just had a couple of questions.

jdbc/src/test/scala/zio/sql/JdbcRunnableSpec.scala
```diff
@@ -55,6 +58,13 @@ trait JdbcRunnableSpec extends ZIOSpecDefault with Jdbc {
       SqlDriver.live
     )
 
+  protected implicit def genInstances[R]
```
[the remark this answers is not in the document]
peixunzhang: Hi,
It is used to generate data for testing here:
https://github.com/zio/zio-sql/pull/695/files#diff-95c65e34d1de6f688c0376b4d84612f8fb80a163383f605194932ac51b427fcfR169

oracle/src/test/scala/zio/sql/oracle/OracleSqlModuleSpec.scala
peixunzhang: Hi,
if I remember correctly there was some issue with reading the data back. Can't check right now because of m1 chip.
Long term we should definitely enable this test.

jczuchnowski (approved):

PR #696, opened 2022-05-27
673 added example and test for union function

### LINKED ISSUE #673: Add examples for union/unionAll
In the module `examples` add some instances for queries using `union` and `unionAll`.
Should be enough to include one example each.

ujali:
Resolves #673


sviezypan (approved):
Thank you!

PR #697, opened 2022-05-27
Added Date Trunc for Postgres

walesho:


postgres/src/main/scala/zio/sql/postgresql/PostgresSqlModule.scala
```diff
@@ -308,6 +308,7 @@ trait PostgresSqlModule extends Sql { self =>
     val BitLength                   = FunctionDef[String, Int](FunctionName("bit_length"))
     val Pi                          = Expr.FunctionCall0[Double](FunctionDef[Any, Double](FunctionName("pi")))
     val GenRandomUuid               = Expr.FunctionCall0[UUID](FunctionDef[Any, UUID](FunctionName("gen_random_uuid")))
+    val DateTrunc                   = FunctionDef[(String, Instant), LocalDateTime](FunctionName("date_trunc"))
```
baldram, line 311: Have you considered using the `ZonedDateTime`?
Looking at spec, this function supports the timezone.
https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
```diff
@@ -1090,6 +1090,34 @@ object FunctionDefSpec extends PostgresRunnableSpec with DbSchema {
...
+    test("date-trunc woth hour") {
+      val someInstant = Instant.parse("1997-05-07T10:15:30.00Z")
+      val query       = select(DateTrunc("hour", someInstant))
+      val testResult  = execute(query)
+
+      val assertion =
+        for {
+          r <- testResult.runCollect
+        } yield assert(timestampFormatter.format(r.head))(
+          Assertion.matchesRegex("[0-9]{4}-[0-9]{2}-[0-9]{2} 10:00:00.0000")
```
baldram, line 1103: What is the reason for using the regex to match the date part here?

walesho: Initially, I compared the actual dates/time. But it was offset by 1hr. 
I assumed it was Timezone difference between the Test-Container used for the test and my JDK. 
Noticed some other test in the codebase used regex so I went with it.

postgres/src/main/scala/zio/sql/postgresql/PostgresSqlModule.scala
walesho: No problem using ZonedDateTime. I could refactor the code to use ZonedDateTime.

baldram: Probably, I would go for `OffsetDateTime` but for consistency I also used the "zoned" one when implementing the `Now()` function.

@jczuchnowski @sviezypan I looked at existing implementation, and other methods are around ZonedDateTime. However, I'm not sure whether, in the DB context, it's not better to use OffsetDateTime. That's my experience from different projects, and also [here is an article](https://www.baeldung.com/java-zoneddatetime-offsetdatetime#differences) that also confirms that the OffsetDateTime will be what people will use in a DB related data model.

If you see it as reasonable, this might be a topic for a separate issue to align all functions.

postgres/src/test/scala/zio/sql/postgresql/FunctionDefSpec.scala
baldram: But I see the time part is constant in this matching so probably the date one even more. I guess the regex is not needed here. 

> Noticed some other test in the codebase used regex so I went with it.

Thos could be about kinda random things like "now" or "current time". Here a test is matching the exact data and exact time, so a straightforward matching shall be sufficient.

postgres/src/main/scala/zio/sql/postgresql/PostgresSqlModule.scala
jczuchnowski: OffsetDateTime does look like a better solution.

jczuchnowski (approved):

PR #699, opened 2022-05-28
Remove initial Literal in Where clause for multiple expression cases

jczuchnowski:


sviezypan (commented):
What about MySQL and Postgres? I think all the dbs have Expr.Literal(true) as a where default.

jczuchnowski:
Split this PR into #711 #712 #713 #714

Issue #707, opened 2022-06-08
Add support for Sum(BigDecimal)  function


balanka:
Like the Sum(Double) function for aggregating columns of Double,
adding this function would allow aggregating columns of BigDecimal
Should we add Avg(BigDecimal) too?

balanka:
I am working on  this issue and have a PR #707 ready for review

PR #751, opened 2022-10-28
#707 Added support for SumInt(Int), SumDec(BigDecimal) and AvgDec(BigDecimal)

balanka:
Added code to support for SumInt(Int), SumDec(BigDecimal) and AvgDec(BigDecimal)

### DIFF
--- a/core/jvm/src/test/scala/zio/sql/ExprSpec.scala
+++ b/core/jvm/src/test/scala/zio/sql/ExprSpec.scala
@@ -1,0 +1,17 @@
+package zio.sql
+
+import zio.test.Assertion.anything
+import zio.test.{ZIOSpecDefault, assert}
+
+object ExprSpec extends ZIOSpecDefault {
+  import ProductSchema._
+
+  def spec = suite("Aggregate Expression")(
+    test("+ works on integer, Double and BigDecimal columns") {
+      val query = agSelect
+      assert(query)(anything)
+    }
+
+  )
+}
+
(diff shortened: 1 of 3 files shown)


core/jvm/src/test/scala/zio/sql/ExprSpec.scala
```diff
@@ -0,0 +1,17 @@
...
+import zio.test.Assertion.anything
+import zio.test.{ZIOSpecDefault, assert}
+
+object ExprSpec extends ZIOSpecDefault {
+  import ProductSchema._
+
+  def spec = suite("Aggregate Expression")(
+    test("+ works on integer, Double and BigDecimal columns") {
+      val query = agSelect
+      assert(query)(anything)
```
sviezypan, line 12: not sure what you're testing here....what's `agSelect` ?

sviezypan (changes requested):
hey @balanka the direction is good but I would like to see a real test against Postgresql or other DB. Also it seems that the build has failed.

core/jvm/src/test/scala/zio/sql/ExprSpec.scala
balanka: Added new enhanced Test AgregationSpec for Postgesql

balanka: The CI-build seem to be flaky my local build was successful
```37 tests passed. 0 tests failed. 1 tests ignored.
Executed in 6 m 15 s```

sviezypan (approved):

PR #771, opened 2022-11-18
Support user defined data types when describing table schema.

sviezypan:
This PR introduces new way to describe tables with the `defineTable[A]` method.

Old way to define table schema was by using DSL and ++ operator.

```
val customers =
      (uuid("id") ++ localDate(“date_of_birth”) ++ string("first_name") ++ string("last_name") ++ boolean("verified_customer") ++ zonedDateTime("created")).table("customers")

    val (customerId, dob, fName, lName, verified, createdString, createdTimestamp) = customers.columns
```

While composable, it was a bit boilerplate-y and another DSL to learn. I think that for Scala devs its more natural to write a case class, which could be then reused when doing select or insert with the same table.

```
  case class Customer(id: UUID, dob: LocalDate, firstName: String, lastName: String, verified: Boolean, 
createdTimestampString: String, createdTimestamp: ZonedDateTime)
```

We can get table description just by calling `defineTable[Customer]` method. The case class data type also needs an implicit schema  (which is required for inserts anyway)

```
    implicit val custommerSchema = DeriveSchema.gen[Customer]

    val customers = defineTable[Customer]

    val (customerId, dob, fName, lName, verified, createdString, createdTimestamp) = customers.columns
```

There are 2 overrides 
`defineTableSmart[Customer]` and `defineTable[Customer]("customers")`, where the first one smartly tries to make plural out of case class name, while taking into account also irregular english plural words.
`OrderOrigin -> order_origins`
`PersonAddress -> person_addresses`
`Foot -> feet`
`defineTable[Customer]("customers")` takes table name explicitly. 
Field names are also converted from camelCase to snake_case
`firstName -> first_name`

There is also a macro which verifies that data type can be used to describe table ( sql does not know sum types). For example:
```
    sealed trait PaymentMethod
    case object Card extends PaymentMethod
    case object Cash extends PaymentMethod

    case class Bill(amount: Int, paymentMethod: PaymentMethod)
    
    val billTable = defineTable[Bill]
```

`defineTable[Bill]` will emit an compilation error 
`Unsupported types by SQL OrderDetails.PaymentMethod`

@jdegoes @jczuchnowski let me know what you think. Docs are updated as well.


core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -4,6 +4,7 @@ import com.github.ghik.silencer.silent
 
 import java.time._
 import scala.language.implicitConversions
+import java.math.BigDecimal
```
jczuchnowski, line 7: doesn't seem to be needed

core/jvm/src/test/scala/zio/sql/ProductSchema.scala
```diff
@@ -10,17 +13,24 @@ object ProductSchema {
...
   import sql._
 
-  val productTable = (
-    string("id") ++
-      localDate("last_updated") ++
-      string("name") ++
-      int("base_amount") ++
-      int("final_amount") ++
-      boolean("deleted")
-  ).table("product")
```
jczuchnowski, line 23: We should still have tests that are using the "old way".

oracle/src/main/scala/zio/sql/oracle/OracleRenderModule.scala
```diff
@@ -447,95 +447,95 @@ trait OracleRenderModule extends OracleSqlModule { self =>
...
             v match {
-              case BigDecimalType                             =>
+              case BigDecimalType                  =>
                 builder.append(value)
                 ()
-              case StandardType.InstantType(formatter)        =>
+              case StandardType.InstantType        =>
                 builder.append(
-                  s"""TO_TIMESTAMP_TZ('${formatter.format(
+                  s"""TO_TIMESTAMP_TZ('${ISO_INSTANT.format(
```
jczuchnowski, line 455: :+1:

oracle/src/test/scala/zio/sql/oracle/OracleSqlModuleSpec.scala
```diff
@@ -129,132 +104,131 @@ object OracleSqlModuleSpec extends OracleRunnableSpec with ShopSchema {
...
-          // result <- execute(read).runHead
-          _ <- execute(delete)
-          // } yield assert(result)(isSome(equalTo(row)))
-        } yield assertCompletes
-
-      }
-    } @@ samples(1) @@ retries(0) @@ shrinks(0)
+    }
+    // TODO uncomment and fix java.sql.SQLDataException: ORA-01874: time zone hour must be between -15 and 15
+    // oracle won't run on m1 machine
```
jczuchnowski, line 109: So this is a problem only on m1 machine?

docs/deep-dive.md
```diff
@@ -7,18 +7,21 @@ title: "Deep dive"
...
 
-trait TableModel extends PostgresModule {
+trait TableModel extends PostgresJdbcModule {
 
-  import ColumnSet._
+  case class Customer(id: UUID, dob: LocalDate, firstName: String, lastName: String, verified: Boolean, createdTimestampString: String, createdTimestamp: ZonedDateTime)
 
-  val customers =
-      (uuid("id") ++ localDate(“date_of_birth”) ++ string("first_name") ++ string("last_name") ++ boolean("verified_customer") ++ zonedDateTime("created"))
-        .table("customers")
```
jczuchnowski, line 19: Later in section `Subqueries & Correlated subqueries` of this document, there are still examples of this kind of definitions but without explanation.
I think we should have both ways well documented. The `++` operator is still useful for defining simple ad-hoc queries for which we don't need to define a model.

oracle/src/test/scala/zio/sql/oracle/OracleSqlModuleSpec.scala
sviezypan: No, the test is failing everywhere but I am not able to reproduce on m1 because of oracle image not running at all.

core/jvm/src/test/scala/zio/sql/ProductSchema.scala
sviezypan: We are not supporting ++ for definition of "source" tables. Let's discuss if there's a use case for having both.

docs/deep-dive.md
sviezypan: As I replaced ++ with data type definition, I now changed it in all places in docs.

jczuchnowski (approved):

PR #791, opened 2022-12-11
More typesafe >, <, = comparisons

sviezypan:
The bug was that compiler did not complain when user compared 2 values of different types:
```
select(customerId, fName, lName, verified, dob).from(customers).where(fName === true)
//or
select(customerId, fName, lName, verified, dob).from(customers).where(verified === "hello")
```

The problem was that in `Expr` where `===` is defined
```
    def ===[F2, A1 <: A, B2 >: B](that: Expr[F2, A1, B2]): Expr[F :||: F2, A1, Boolean] =
      Expr.Relational(self, that, RelationalOp.Equals)
```
When `B =:= Boolean` and `B2 =:= String`, Scala happily infers `B2` to be `Any`, so that would compile.

This PR introduces `ComparableTypes` trait which would prevent that behaviour. 
We still allow to compare different numeric or date types like `3 < 5.5` etc.


core/jvm/src/main/scala/zio/sql/expr.scala
```diff
@@ -461,4 +474,25 @@ trait ExprModule extends NewtypesModule with FeaturesModule with OpsModule {
         def typeTag = implicitly[TypeTag[Value]]
       }
   }
+
+  @implicitNotFound(
+    "You cannot compare values of different types ${A} and ${B}. " +
+      "As those are unrelated types, this query would blow up at database level."
```
jczuchnowski, line 480:
```suggestion
      "As those are unrelated types, this query would fail at database level."
```

jczuchnowski (approved):

Issue #583, opened 2022-02-14
Expr doesn't fold nicely when dynamically constructing query


sviezypan:
Let's say we are constructing dynamic query based on some optional values.

Non dynamic example is:
```
    val persons = (string("name") ++ int("age")).table("persons")
    val (name, age) = persons.columns

    select(name ++ age).from(persons).where(age > 10 && age < 50)
```

Now if we have `min age` and `max age` as dynamic values, it's hard to fold them into final where clause

```
    val minAge: Option[Int] = Some(10)
    val maxAge: Option[Int] = None

    val minAgeQuery = minAge.map(m => age >= m)
    val maxAgeQuery = maxAge.map(m => age <= m)   

    val whereExpr = 
        List(minAgeQuery, maxAgeQuery)
                .flatten
                .fold[Expr[_, persons.TableType, Boolean]](Expr.literal(true))(_ && _)

   // or
    val whereExpr2 = 
        List(minAgeQuery, maxAgeQuery)
            .flatten
            .reduceLeftOption[Expr[_, persons.TableType, Boolean]](_ && _)
            .get

    select(name ++ age).from(persons).where(whereExpr)
```

I would like to construct where clause without specifying `Expr[_, persons.TableType, Boolean]` type. 
Thats not currently possible because && operator combines phantom type Fs of Expr[F, _, _]  into Features.Union.

```
//actual
Expr[Features.Union[Features.Source["age".type], Features.Source["age".type]], metroLine.TableType, Boolean]

//expected
Expr[Features.Source["age".type], metroLine.TableType, Boolean]
```

Probably we need some other operator on Expr that would allow us similar actions, any ideas are welcome.

sviezypan:
The best way to write it is probably something like

```
val minStations: Option[Int] = Some(10)
val maxStations: Option[Int] = None
val sortDesc: Boolean = true
 
val minStationsQuery = minStations.map(m => stationCount >= m).getOrElse(Expr.literal(true))
val maxStationsQuery = maxStations.map(m => trackType <= m).getOrElse(Expr.literal(true))
 
val ord =
       if (sortDesc)
           stationCount.desc
       else
           stationCount.asc

val whereExpr =
	minStationsQuery && maxStationsQuery
```

We need to explore if its possible to make phantom type F in Expr covariant, then this might compose nicer.

PR #813, opened 2023-01-19
Contravariant F and optionals

sviezypan:
The recent changes that started using macros to verify `where`, `inserts`. `group by` etc made possible to get rid of `Features.Union[_, _]` and start using intersection types wherever `F` composes.
So the phantom type `F`  of x in 
`val x = name ++ age`
will be `Features.Source["name"] with Features.Source["age"]` instead of 
`Features.Union[Features.Source["name"], Features.Source["age"]]`

Its possible now to make F contravariant which gives us nice type inference and makes working with dynamic queries nicer. The example of #583 is now possible.
```
// before
List(where1, where2, where3)
     .fold[Expr[_, TableType, Boolean]](Expr.literal(true))(_ && _)

// now
List(where1, where2, where3)
   .fold(Expr.literal(true))(_ && _)
```

This PR also makes changes to working with nullable columns. 
Let's say that `link` describes a nullable column and has a type `Option[String]`. 
Instead of always having to use `Option("some link")`, users have more possibilities
```
      select(cityName).from(city).where(link === Option(""))
      select(cityName).from(city).where(link === Some(""))
      select(cityName).from(city).where(Some("") === link)

      select(cityName).from(city).where(link === None)
      select(cityName).from(city).where(None === link)

      select(cityName).from(city).where(link === "")
      select(cityName).from(city).where(" " === link)
```


macros/src/main/scala-3/zio/sql/macros/wherelike.scala
```diff
@@ -5,6 +5,8 @@ sealed trait WhereIsSound[WhereF, GroupByF]
 object WhereIsSound {
   final case class WhereCanBeCalled[WhereF, GroupByF]() extends WhereIsSound[WhereF, GroupByF]
 
-  implicit def materializeWhereIsSound[WhereF, GroupByF]: WhereIsSound[WhereF, GroupByF] = WhereCanBeCalled[WhereF, GroupByF]()
+  implicit def materializeWhereIsSound[WhereF, GroupByF]: WhereIsSound[WhereF, GroupByF] = {
 
+    WhereCanBeCalled[WhereF, GroupByF]()
+  }
```
jczuchnowski, line 11: I suppose this should go back to the original formatting.

```suggestion
  implicit def materializeWhereIsSound[WhereF, GroupByF]: WhereIsSound[WhereF, GroupByF] = WhereCanBeCalled[WhereF, GroupByF]()
```

jczuchnowski (approved):

PR #867, opened 2023-05-31
Quoted table names and columns should handle dots.

daharon:
This is so that table names in non-default PostgreSQL [schemas](https://www.postgresql.org/docs/current/ddl-schemas.html) like `myschema.mytable` will be rendered with quotes properly, eg. `"myschema"."mytable"`

This came about because I have a table defined like so:
```scala
val user = defineTable[User]("accounts.user")
```
But when the query is rendered it ends up like below, which doesn't parse correctly in PostgreSQL:
```sql
SELECT "accounts.user"."id", "accounts.user"."email" FROM "accounts.user"
```
With this change the query will be rendered like so:
```sql
SELECT "accounts"."user"."id", "accounts"."user"."email" FROM "accounts"."user"
```


daharon:
> This fails for Scala 2.12 due to the limitation in String interpolation [scala/bug#6476](https://github.com/scala/bug/issues/6476). Changing to String concatenation should help.

Thanks, I updated it.

daharon:
There's one failing test:
https://github.com/zio/zio-sql/actions/runs/5179946353/jobs/9335238448?pr=867

But it doesn't seem to be related to my changes.
```
[info] compiling 3 Scala sources to /home/runner/work/zio-sql/zio-sql/postgres/target/scala-2.13/classes ...
[error] ## Exception when compiling 9 sources to /home/runner/work/zio-sql/zio-sql/core/jvm/target/scala-2.13/test-classes
[error] java.lang.StackOverflowError
```

jczuchnowski (approved):

jczuchnowski:
Finally solved the stack overflow problem.

Thanks for the contribution!

PR #876, opened 2023-06-08
Make null primitive value return None

tonykwok1992:
Currently selecting rows with null value in a nullable int/long columns will gives Some(0) even when the schema object field has type Option[Int], due to the fact that resultSet.getInt(columnIndex) gives 0 for null value rather than null,


postgres/src/test/scala/zio/sql/postgresql/SelectSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+    test("Select null int value as None Option") {
+      val movieColumns = MoviesSchema.id ++ MoviesSchema.rating
+      val selectStmt = select(movieColumns).from(MoviesSchema.movies).where(MoviesSchema.id === 1)
+      val selectResult = execute(selectStmt.to((MoviesSchema.Movies.apply _).tupled)).runCollect
+
+      val insertAssertion = for {
+        movies <- selectResult
+      } yield assert(movies.toList)(equalTo(List(MoviesSchema.Movies(
+        id = 1,
+        rating = None
```
tonykwok1992, line 20: This is Some(0) before the change

tonykwok1992:
Hello, can I get a review on this? thanks!

postgres/src/test/scala/zio/sql/postgresql/SelectSpec.scala
```diff
@@ -0,0 +1,32 @@
+package zio.sql.postgresql
+
+import zio.Cause
+import zio.test.Assertion._
+import zio.test._
+
+object SelectSpec extends PostgresRunnableSpec with DbSchema {
```
sviezypan, line 7: would you move this to PostgresSqlModuleSpec? as there are also tests for selects

postgres/src/test/scala/zio/sql/postgresql/SelectSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+          equalTo(
+            List(
+              MoviesSchema.Movies(
+                id = 1,
+                rating = None
+              )
+            )
+          )
+        )
+      insertAssertion.mapErrorCause(cause => Cause.stackless(cause.untraced))
```
sviezypan, line 28: I think it would work if you just return the for comprehension directly?

sviezypan (changes requested):
hi @tonykwok1992 thank you for this work, this is great catch and fix. Just 2 minor comments and then its good to be merged.

PR #987, opened 2024-04-28
DRAFT: feat: implement sounds like function for mysql

GrigoriiBerezin:


GrigoriiBerezin:
I'm not sure where it should be - in core or in MySql module, but I believe it should be separate from core module

